PR #65, opened 2020-09-28
cip: IdentityIndex

oed:
PR for CIP 11
Discussions: #3


PaulLeCam (approved):
Looks great!

CIPs/CIP-11/CIP-11.md
```diff
@@ -0,0 +1,261 @@
...
+For encrypted data sets the *key* is simply a multibase (base32) encoded string of 24 random bytes. The *value* is the DocID of the *reference* document with content encryption. See the [Data Set Keychain Definition](#data-set-keychain-definition) section for more more details on how this works.
+
+#### Schema
+
+**Deployment:** `<idx-schema-DocID>`
+
+```json
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "IDX",
```
PaulLeCam, line 69: `IdentityIndex` instead of `IDX`

oed: fixed

CIPs/CIP-11/CIP-11.md
```diff
@@ -0,0 +1,261 @@
...
+
+The *definition-reference* pair can be stored publicly or encrypted. The former is desirable for data sets which require public discoverability. The latter is useful when possession of the data set should be kept secret. The encrypted *reference* should be encrypted using a unique symmetric key. This enables the data set to be selectively disclosed by asymmetrically encrypting the symmetric key to a given recipient.
+
+![IDX Example](./assets/idx-example.png)
+
+### IDX Document
+
+The IDX document simply consist of a map from strings to DocIDs. For public data sets the *key* in this map is the DocID (without the prepended `ceramic://`) of the *definition* document, and the *value* is the DocID (with `ceramic://`) of the *reference* document.
```
PaulLeCam, line 58: I think we should enforce validation in the schema to avoid possibly invalid keys here.

The nice thing if both DocIDs and encrypted data keys are encoded in base 36 is that we can validate based on this.
We should also use a multicodec when setting an encrypted data key (could be simply `raw`), this way we can easily differentiate from DocIDs.

I think the logic would be something like:
1. Regexp matching base36 encoding byte followed by valid base 36 characters
1. Check second byte:
    - if matching DocID multicodec, it's not encrypted
    - if matching raw multicodec, follow encrypted logic
    - otherwise it's an invalid key per IDX spec

oed: I don't think it's that easy to validate base36 using regexp unfortunately. We could add a max length and enforce that only `[0-9][a-z]` is used however. Would want to see some examples from the DocID implementation before deciding on length though.

PaulLeCam: It seems that `f` is the character used for base 16 so I imagine we can have a regexp like `/^fa|b[0-9a-z]+$/` (taking `a` and `b` as example codes for DocID or raw bytes) to validate the keys, no?

oed: `base36` is actually `k`. Might be possible to do this, but we need examples from the implementation first.

PaulLeCam: As I'm going through the necessary changes in the IDX lib implementation I think we really need to have a way to validate these keys so that IDX can throw useful errors when invalid, otherwise it will be way harder to properly catch possible issues and implement the necessary logic for encrypted data keys.

oed: Not sure i follow why it would be difficult to catch errors? If DocID decoding fails, then check check if it's a base36 encoded string?

PaulLeCam: Well we go from having one way to identify a document (the `ceramic://...` URL) to 3 (URL, string or encoded key) with possible associated logic and conversions all over the place, for example the Ceramic interactions need the URL form but the index keys should use the string form, so I think we need to at least have a good way to identify the inputs and outputs at the edges.

oed: So only `ceramic://` should be a valid "link" to another ceramic document. In the *key* of the map we can't really have links anyway and if you ever where to specify a path, e.g. `<idx-docid>/<def-docid>/<prop-in-reference>` it wouldn't really make sense to include `ceramic://`. 
The random key is not really a way to identify a definition, rather a way to find a definition that you've decrypted from the keychain.

This is nuanced though. Let's maybe find some time for a live sync?

PaulLeCam: Yes I think we need to define the expected behavior from the developer side, what would the APIs look like in these different cases? For example:
- `idx.get('profile', aliceDID)`
- `idx.set(profileDefinitionId, { name: 'Bob' })`
- `idx.set('my-encrypted-data', { data: cleartext })`: can we do that? what is the expected behavior for the library? is the encryption based on the matching definition for 'my-encrypted-data'?
- `idx.set('profile',  { name: 'Bob' }, { encryptedFor: [aliceDID] })`: can we do that? what if 'profile' is already set but not encrypted?
- `idx.get('random app key')`: need to be able to clearly state if the key is a definition alias, definition ID or encrypted key

oed:
@PaulLeCam I guess we should change the schema to not include `ceramic://` for the values?

PaulLeCam:
Yes good idea, maybe we can refer to these as "DocID string" or another term we could use for these?

Pinkipunkki:
I’ve never been on GitHub before. This is my first post. I do wish to be a user of web3. I have recently dumped my surname(s) so my idea is - my dad said to me once: what is wrong with your real name? There is nothing wrong with my real name I thought back then, but now I would want to add to it. My first names are ‘anita jane’ on web3 in terms of a ‘root server’ I would like to be known as ‘anita Jane Jane’ after all web3, is web and number 3 - so take it literally - 3 names. Just a thought. I have an artist name: pinkipunki, then I remastered this name an added an extra K - pinkipunkki, I could go further and do pinkipunkkki and it would still be understood. The secret is in the 3. Less is always more when you flip it on it’s head.

Issue #32, opened 2020-05-26
Discussion: Basic Profile Definition


michaelsena:
Discussion for CIP-19: Basic Profile Definition.
Up to date spec in PR https://github.com/ceramicnetwork/CIP/pull/66

oed:
A json-schema for the basic profile could look something like this:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
    "$ref": "#/definitions/BasicProfile",
    "definitions": {
      "BasicProfile": {
        "title": "BasicProfile",
        "type": "object",
        "properties": {
          "name": {
            "type": "string",
            "title": "name",
            "maxLength": 150

          },
          "image": {
            "type": "string",
            "title": "image"
          },
          "description": {
            "type": "string",
            "title": "description",
            "maxLength": 420
          },
          "emoji": {
            "type": "string",
            "title": "emoji",
            "maxLength": 2
          },
          "background": {
            "type": "string",
            "title": "background"
          },
          "birthDate": {
            "type": "integer",
            "title": "birthDate"
          },
          "url": {
            "type": "string",
            "title": "url",
            "maxLength": 240
          },
          "gender": {
            "type": "string",
            "title": "gender",
            "maxLength": 42
          },
          "homeLocation": {
            "type": "string",
            "title": "homeLocation",
            "maxLength": 140
          },
          "nationality": {
            "type": "string",
            "title": "nationality",
            "maxLength": 140
          },
          "affiliation": {
            "type": "array",
            "items": {
              "type": "string",
              "maxLength": 140
            },
            "title": "affiliation"
          }
        }
      }
    }
}
```

michaelsena:
@oed I assume the "title": "affiliation" should be "title": "BasicProfile" , correct?

PaulLeCam:
I think as much as possible these values should match common standards to make them easier to implement, especially across apps that may use different locales, for example:

- `gender` could implement [ISO 5218](https://en.wikipedia.org/wiki/ISO/IEC_5218)
- `nationality` could use [ISO 3166](https://en.wikipedia.org/wiki/ISO_3166) (maybe even be strict about using alpha-2 or alpha-3 variant)
- `birthDate` shouldn't be a UNIX timestamp because that wouldn't work for people born before 1970, instead it could use the `YYYY-MM-DD` format of [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

[...]

PR #66, opened 2020-09-29
cip: Basic Profile Definition

oed:
PR for CIP-19: Basic Profile Definition
Discussions: https://github.com/ceramicnetwork/CIP/issues/32

### DIFF
--- a/CIPs/CIP-19/CIP-19.md
+++ b/CIPs/CIP-19/CIP-19.md
@@ -112,26 +112,18 @@ The Basic Profile schema defines the format of a document that contains the prop
       "pattern": "^[A-Z]{2}$"
     },
     "nationalities": {
-      "oneOf": [
-        {
-          "type": "string",
-          "pattern": "^[A-Z]{2}$"
-        },
-        {
-          "type": "array",
-          "minItems": 1,
-          "items": {
-            "type": "string",
-            "pattern": "^[A-Z]{2}$"
-          }
-        }
-      ]
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "type": "string",
+        "pattern": "^[A-Z]{2}$"
+      }
     },
     "affiliations": {
       "type": "array",
       "items": {
         "type": "string",
-        "maxLength": 140
+        "maxLength": 240
       }
     }
   }
--- a/CIPs/CIP-19/CIP-19.md
+++ b/CIPs/CIP-19/CIP-19.md
@@ -53,10 +53,10 @@ The Basic Profile schema defines the format of a document that contains the prop
 | Property           | Description                    | Value                                                        | Max Size | Required | Example                      |
 | ------------------ | ------------------------------ | ------------------------------------------------------------ | -------- | -------- | ---------------------------- |
 | `name`             | a name                         | string                                                       | 150 char | optional | Mary Smith                   |
-| `image`            | an image                       | IPFS CID                                                     |          | optional |                              |
+| `image`            | an image                       | IPFS CID URI                                                 |          | optional |                              |
 | `description`      | a short description            | string                                                       | 420 char | optional | This is my cool description. |
 | `emoji`            | an emoji                       | unicode                                                      | 2 char   | optional | 🔢                            |
-| `background`       | a background image (3:1 ratio) | IPFS CID                                                     |          | optional |                              |
+| `background`       | a background image (3:1 ratio) | IPFS CID URI                                                 |          | optional |                              |
 | `birthDate`        | a date of birth                | [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)           | 10 char  | optional | 1990-04-24                   |
 | `url`              | a url                          | string                                                       | 240 char | optional | http://ceramic.network       |
 | `gender`           | a gender                       | string                                                       | 42 char  | optional | female                       |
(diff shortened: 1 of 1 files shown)


CIPs/CIP-19/CIP-19.md
```diff
@@ -0,0 +1,180 @@
...
+    "homeLocation": {
+      "type": "string",
+      "maxLength": 140
+    },
+    "residenceCountry": {
+      "type": "string",
+      "pattern": "^[A-Z]{2}$"
+    },
+    "nationalities": {
+      "oneOf": [
```
PaulLeCam, line 115: Is think we should make it so it's always an array, rather than having to check at runtime for an array or scalar.

CIPs/CIP-19/CIP-19.md
```diff
@@ -0,0 +1,180 @@
...
+| `birthDate`        | a date of birth                | [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)           | 10 char  | optional | 1990-04-24                   |
+| `url`              | a url                          | string                                                       | 240 char | optional | http://ceramic.network       |
+| `gender`           | a gender                       | string                                                       | 42 char  | optional | female                       |
+| `residenceCity`    | a city of residence            | string                                                       | 140 char | optional | Berlin                       |
+| `residenceCountry` | a country of residence         | [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) | 2 char   | optional | DE                           |
+| `nationalities`    | nationalities                  | array of [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) values | 40 char  | optional | CN                           |
```
PaulLeCam, line 65: I don't think the 40 char limit makes sense here, or at least I don't see how it's validated by the schema?

CIPs/CIP-19/CIP-19.md
```diff
@@ -0,0 +1,180 @@
...
+| `affiliations`     | affiliations                   | array of strings                                             | 240 char | optional | Ceramic Ecosystem Alliance   |
```
PaulLeCam, line 66: Spec here states 240 char limit but the JSON schema limits to 140 char per item.

CIPs/CIP-19/CIP-19.md
oed: Should it be validated?  what should the limit be?

CIPs/CIP-19/CIP-19.md
oed: I'll update the schema!

CIPs/CIP-19/CIP-19.md
oed: Good point!

CIPs/CIP-19/CIP-19.md
PaulLeCam: The items in the array should be limited to 2 chars length, and maybe we could limit the number of items in the array?

oed: updated now

PaulLeCam (approved):

CIPs/CIP-19/CIP-19.md
```diff
@@ -0,0 +1,173 @@
...
+  "title": "BasicProfile",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "maxLength": 150
+    },
+    "image": {
+      "type": "string"
+    },
```
oed, line 82: @PaulLeCam thoughts about changing this to be a string that starts with `ipfs://`? Same thing for background.

PaulLeCam: I think we should make the type more strict but limiting to `ipfs://` might limit use cases. We'd likely also need to include `ipns://`, but what about other protocols and possible browser extensions (ex `web+ipfs://`)?
Maybe instead we can restrict it to the [`uri` format](https://json-schema.org/understanding-json-schema/reference/string.html#uri-template)?

oed: I'd rather be restrictive here so that developers know what to expect. This will in most cases need to be handled by developers in their app, so the fewer options the better imo.

PaulLeCam: Yes I agree it's a good thing to be more restrictive here, but I feel IPFS only is very restrictive and might limit usage for projects that would otherwise use this definition. If we want to limit to a single protocol I think `https://` would be an more convenient one for most projects, at least in the short term.

oed: Don't agree at all tbh. The profiles should be trustless. Using HTTPS opens up attack vectors which I'm definitely not comfortable with. Ipfs is extremely descriptive because of the content addressing. I think immutability is what we should aim for here. This is critical for a lot of initial use cases of IDX where trust is extremely important, e.g. decentralized governance.

PaulLeCam: Yes I totally agree this is the "web3 ideal" and we should definitely aim for this, but as we're trying to figure out the "web 2.5 bridge" we should also consider what this means in terms of UX and DX.
Currently is that it's not really possible to be fully decentralized, we're still using gateways for IPFS, hosting on AWS, etc. so we're already making these trade-offs at a different layer. By forcing to use IPFS here, I'm concerned we're "forcing" users and dev to the opt for the most secure option, even though we're still relying on centralized options ourselves.
I think there is a risk we will limit the potential use case for the basic profile definition and that it could be used as a "universal default" if we make it too constraining. Developers could simply use an alternative schema allowing HTTPS that would "compete" with our basic profile, and split the ecosystem rather than helping it converge to a single go-to option.
Maybe we could discuss this with other people in the ecosystem to make sure we base these kind of decisions on current product needs rather than decentralizations ideals?

oed: Most of our initial customers care deeply about this and if we start getting competing basic profiles in IDX that's actually great because we get more usage of IDX.

oed: Using gateways for ipfs still provide much more guarantees because of content adressability, and we also have the option to upgrade very easily to more secure alternatives (once we start running ipfs and ceramic in the browser).

Issue #44, opened 2020-06-15
RFC: Identity Claims Document


michaelsena:
---
cip: 
title: Identity Claims Document
author: Michael Sena (@michaelsena)
discussions-to:
status: Idea
category: Standards
type: RFC
created: 2020-06-15
requires:
replaces:
---
  
## Simple Summary

A document that stores links to various claim documents that make statements about a DID's identity.

### Document

Doctype: Tile

### Definitions

`provider`:

`claims`:

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

oed:
The schema for this is a bit tricky. There is a regex that matches the key and the value of the map:
```json
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "$ref": "#/definitions/AccountLinks",
  "definitions": {
    "AccountLinks": {
      "type": "object",
      "patternProperties": {
        "^[a-zA-Z0-9]{1,63}@[-a-zA-Z0-9]{3,16}:[-a-zA-Z0-9]{1,47}": {
            "type": "string",
            "pattern": "^ceramic:\/\/.+"
        }
      },
      "additionalProperties": false,
      "title": "AccountLinks"
    }
  }
}
```

PR #67, opened 2020-09-30
cip: Crypto Accounts Definition

oed:
PR for CIP-21
Discussion: https://github.com/ceramicnetwork/CIP/issues/44


PaulLeCam (approved):
LGTM, just wondering if we should use the same name of "Crypto Accounts" or "Crypto Account Links" for both the definition and schema, or it's better to use different names as in the spec?

oed:
My bad @PaulLeCam, forgot to change it everywhere. We should only use **Crypto Accounts**. Should be updated now!

Issue #55, opened 2020-07-22
Linked Accounts Array


michaelsena:
```
cip: 
title: Linked Accounts Array
author: Michael Sena (@michaelsena), Joel THorstensson (@oed)
discussions-to:
status: Idea
category: Standards
type: RFC
created: 21-07-2020
requires: Tile Doctype (CIP-8)
replaces:
```
## Simple Summary
**Linked Accounts Array** defines the schema for a document that stores a list of a DID's linked accounts.

## Abstract
The Linked Accounts Array is used to build public links between a DID and their various domain names, social accounts, or other DIDs. The accounts contained in this document are not able to authenticate or control the DID, they are merely public associations. Documents that implement this schema can easily create documents that maintain these lists in a standardized way.

This document is capable of storing any account that can publish a message including a DID. Although blockchain/crypto externally owned accounts and contracts can technically be stored in this document, for those use cases it is recommend that you instead use a [Crypto Directory (CIP-21)](https://github.com/ceramicnetwork/CIP/issues/44) document which implements a schema specialized for this use case.

## Motivation
Such a standard would be helpful in defining a schema that many in the Ceramic ecosystem can use when creating documents that seek to associate third-party accounts with a DID.

## Specification

The Linked Accounts Array specification consists of a doctype, schema, and tags.

### Doctype

The Linked Accounts Array is stored in a [Tile doctype (CIP-N)](https://github.com/ceramicnetwork/CIP/issues/9).

### Schema

The Linked Accounts Array schema defines a document which maintains an array of JSON objects that represent a DID's linked accounts. Below find the fields that are contained in each object.

- `protocol`: The network protocol used to resolve the host. A string. Required. (i.e. http)
- `host`: The host of the ID. A string. Required. (i.e. github.com)
- `id`: The path of the user's unique ID within the host. Can be the same as the host to represent the root. A string. Required. (i.e. github.com/marysmith)
- `claim`: The unique location where the DID string is stored. The location should be a place where only the given ID could have posted the DID. A string. Required. (i.e. gist.github.com/marysmith/5c48debdb7089b3c8f86cca31739572c)
- `proofs`: Links to proofs issued by third-party services which have verified the claim against the username. An array of strings. Optional. (i.e. ceramic://bafy...123)

```
TODO: formalize the JSON schema
```

### Tags
When creating a document that uses the Linked Accounts Array schema, add `LinkedAccountsArray` to the **tags** field in the document header.

## Examples

[...]

PR #70, opened 2020-11-03
cip: Also Known As Definition

oed:
PR for CIP-23
Discussion: https://github.com/ceramicnetwork/CIP/issues/55

### DIFF
--- a/CIPs/CIP-23/CIP-23.md
+++ b/CIPs/CIP-23/CIP-23.md
@@ -116,7 +116,7 @@ In the config of the *definition* defined above we can see a `tags` property. We
 
 ### Example
 
-An example Crypto Accounts *reference* document that includes two Ethereum accounts and one Bitcoin account.
+An example Social Accounts *reference* document that includes two accounts, Twitter and Github.
 
 ```js
 const profile = await ceramic.createDocument('tile', {
@@ -145,7 +145,7 @@ const profile = await ceramic.createDocument('tile', {
 
 ## Suggested Usage
 
-It is recomended to always include both the `claim` and the `attestation` property. The claim is usually a way to verify that the social account actually want's to be associated with the users DID. The attestation is a [Verifiable Credential](https://github.com/decentralized-identity/did-jwt-vc) from a third party that this is the case. In some scenarious it might be impossible to provide a publicly verifiable `claim`. In these cases just an `attestation` will have to suffice.
+It is recomended to always include both the `claim` and the `attestation` property. The claim is usually a way to verify that the social account actually want's to be associated with the users DID. The attestation is a [Verifiable Credential](https://github.com/decentralized-identity/did-jwt-vc) from a third party that this is the case. In some scenarious it might be impossible to provide a publicly verifiable `claim`, for example it's not possible to create a public post which can be linked to in an easy. In these cases just an `attestation` will have to suffice. Note however that this means that there is full trust in the issuer of this attestation, where as when there is a public claim it's always possible to double check the truth of the attestation.
 
 
 ## Rationale
@@ -156,7 +156,7 @@ It is recomended to always include both the `claim` and the `attestation` proper
 ## Implementation
 
 - [**js-idx**](https://idx.xyz/): A javascript library for reading and writing data sets associated with a DID
-- [**js-idx-constants**](https://github.com/ceramicstudio/js-idx-constants): A javascript library to be used with js-idx which contains the Basic Profile definition.
+- [**js-idx-constants**](https://github.com/ceramicstudio/js-idx-constants): A javascript library to be used with js-idx which contains the Social Accounts definition.
 
 
 ## Copyright
--- a/CIPs/CIP-23/CIP-23.md
+++ b/CIPs/CIP-23/CIP-23.md
@@ -1,0 +1,164 @@
+```
+cip: 23
+title: Social Accounts Definition
+author: Michael Sena (@michaelsena), Joel Thorstensson (@oed)
+status: Last Call
+category: Standards
+type: RFC
+created: 2020-07-21
+edited: 2020-11-03
+requires: Tile Doctype (CIP-8)
+```
+
+## Simple Summary
+
+**Social Accounts** is a data set that stores a list of a social accounts that are publicly linked to the users DID.
+
+
+## Abstract
+
+Users may want to publicly associate any number of social accounts, such as Twitter, GitHub, discord, etc. to their DID. Creating and storing public verifications for these various accounts in the Social Accounts *reference* document enables anyone to observe these associations and as a result infer trust about the given user. 
+
+Adding an account to the Social Accounts document *does not* allow that account to control or authenticate the DID. Rather it is simply just a verifiable public mapping.
+
+
+## Motivation
+
+Publicly associating one or more social accounts to a DID offers a number of benefits.
+
+- Public association with a DID to any number of accounts
+- Serves as a foundation for building aggregated reputation on an identity comprised of multiple accounts, not individual one-off accounts
+- Provides a common interface for creating a list of linked social accounts, making discoverability and query predictable
+
+
+## Specification
+
+The Social Accounts is a *definition* that contains an array of *linked accounts* entries.
+
+### Definition content
+
+**Deployment:** `<definition-DocID>`
+
+```json
+{
+  "name": "Social Accounts",
+  "description": "Social accounts linked to your DID",
+  "schema": "<reference-schema-DocID>",
+  "config": {
+    "tags": ["SocialAccounts"]
+  }
+}
+```
+
+### Reference Schema
+
+The reference schema defines a document which maintains an array of JSON objects that represent a DID's linked accounts. Below find the fields that are contained in each object.
+
+| Property       | Description                                                  | Value            | Max Size  | Required | Example                                      |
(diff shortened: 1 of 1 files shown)


CIPs/CIP-23/CIP-23.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Reference Schema
+
+The reference schema defines a document which maintains an array of JSON objects that represent a DID's linked accounts. Below find the fields that are contained in each object.
+
+| Property       | Description                                                  | Value            | Max Size  | Required | Example                                      |
+| -------------- | ------------------------------------------------------------ | ---------------- | --------- | -------- | -------------------------------------------- |
+| `protocol`     | Network protocol used to resolve the host                    | string           | 50 char   | required | https                                        |
+| `host`         | Host used to resolve the ID                                  | string           | 150 char  | required | https://github.com                           |
+| `id`           | Path of the unique ID within the host. Can be the same as the host to represent the root. | string           | 450 char  | required | https://github.com/marysmith                 |
```
PaulLeCam, line 61: What is the rationale between the `protocol`, `host` and `id` separation please?
It seems the `id` includes all the information provided by the other fields anyways. Maybe we could make it a `url` field with the associated validation requirement to be more explicit?

oed: I think it was made this way to accommodate for other types of social accounts on a variety of different protocols. This proposal was previously more broad. I guess we can simplify it. I think `url` makes sense.
Would be good to get input from @michaelsena here as well

michaelsena: We made this general to support additional account types beyond Web2 social. In the case of a Matrix ID, or a Web3 domain name, you'd need additional data to specify where the registry is that might not fit into just a URL.

In the example of matrix:
Protocol: Matrix
Host: the location of your homeserver
ID: Your unique user ID on that homeserver (this is not an /append to the host as is the case with twitter)

In the example of ENS:
Protocol: Ethereum
Host: The chainID + contract address
ID: The user domain (or subdomain)

I'm wondering if we broaden this definition to be something like "Also Known As" which includes all of the linked accounts of the user. This is the terminology used by W3C in the DID spec nowadays.

With this naming though, I'm just wondering how we differentiate between Also Known As, and Crypto (CAIP-10) Accounts.

CIPs/CIP-23/CIP-23.md
```diff
@@ -0,0 +1,164 @@
...
+}
+```
+
+### Tags
+
+In the config of the *definition* defined above we can see a `tags` property. We simply apply the tags in the config to the `tags` property in the *metadata* of the *reference* Tile.
+
+### Example
+
+An example Crypto Accounts *reference* document that includes two Ethereum accounts and one Bitcoin account.
```
michaelsena, line 119: This should not be crypto accounts. This should describe what's included in this definition.

CIPs/CIP-23/CIP-23.md
```diff
@@ -0,0 +1,164 @@
...
+		claim: "https://gist.github.com/marysmith/5c48debdb7089b3c8f86cca31739572c"
+		attestations: ["eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9..."]
+	}]
+})
+```
+
+
+## Suggested Usage
+
+It is recomended to always include both the `claim` and the `attestation` property. The claim is usually a way to verify that the social account actually want's to be associated with the users DID. The attestation is a [Verifiable Credential](https://github.com/decentralized-identity/did-jwt-vc) from a third party that this is the case. In some scenarious it might be impossible to provide a publicly verifiable `claim`. In these cases just an `attestation` will have to suffice.
```
michaelsena, line 148: I think we could make the last sentence more descriptive by saying that we can include the ID of a non-public post, but in these cases an attestation is required and there will be more trust in the attestation since it's not a publicly verifiable claim. (or something along those lines)

CIPs/CIP-23/CIP-23.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Rationale
+
+**Decentralization & Trust:** Linked social account directories are data that needs to be globally-available, cross-chain, censorship-resistant, and live permissionlessly in the public domain (not on any single server). Additionally this information should be owned by a DID and will need to be updated from time to time. These requirements make Ceramic the most appropriate platform for publishing account links content.
+
+
+## Implementation
+
+- [**js-idx**](https://idx.xyz/): A javascript library for reading and writing data sets associated with a DID
+- [**js-idx-constants**](https://github.com/ceramicstudio/js-idx-constants): A javascript library to be used with js-idx which contains the Basic Profile definition.
```
michaelsena, line 159: should be "which contains the <insert name we land on> definition." (Not Basic Profile)

michaelsena:
@oed @PaulLeCam I made some comments

CIPs/CIP-23/CIP-23.md
oed: good catch, ty

CIPs/CIP-23/CIP-23.md
oed: I'm wondering if it makes sense to restrict this to mainly web2 accounts. Or alternatively merge it with crypto accounts (but that might cause more confusion).

CIPs/CIP-23/CIP-23.md
oed: Err, yeah this has not been updated!

CIPs/CIP-23/CIP-23.md
```diff
@@ -0,0 +1,176 @@
...
+
+An example Social Accounts *reference* document that includes two accounts, Twitter and Github.
+
+```js
+const profile = await ceramic.createDocument('tile', {
+  metadata: {
+    schema: "<reference-schema-DocID>"
+    tags: ["AlsoKnownAs"]
+  },
+  content: [{
```
stbrody, line 139: Putting this as an array means that if someone wants to look up the twitter handle for a given DID, they have to load this entire document and then iterate over this array to find the twitter entry.  Should this instead be a map keyed by 'host' to make it easier to locate specific entries?

CIPs/CIP-23/CIP-23.md
```diff
@@ -0,0 +1,176 @@
...
+    claim: "https://gist.github.com/marysmith/5c48debdb7089b3c8f86cca31739572c",
+    attestations: [{ did-jwt-vc: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9..." }]
+  }]
+})
+```
+
+
+## Suggested Usage
+
+It is recomended to always include both the `claim` and the `attestation` property. The claim is usually a way to verify that the social account actually want's to be associated with the users DID. The attestation is a [Verifiable Credential](https://github.com/decentralized-identity/did-jwt-vc) from a third party that this is the case. In some scenarious it might be impossible to provide a publicly verifiable `claim`, for example it's not possible to create a public post which can be linked to in an easy. In these cases just an `attestation` will have to suffice. Note however that this means that there is full trust in the issuer of this attestation, where as when there is a public claim it's always possible to double check the truth of the attestation.
```
stbrody, line 160: nit: /s/want's/wants

stbrody, line 160: > In some scenarious it might be impossible to provide a publicly verifiable `claim`, for example it's not possible to create a public post which can be linked to in an easy

in an easy...?

Perhaps something like "for example, in some systems it may not be possible to create a public post which can be easily linked to"

CIPs/CIP-23/CIP-23.md
oed: Last time we discussed this you suggested that we put the `host` as the key and value being an array of accounts. Now I'm leaning towards keeping the current design for two reasons:
1. All protocols might not require a "host" (they only use protocol and id)
2. You would still need to load all data, seems like it wouldn't be too much overhead to just iterate though the array.

CIPs/CIP-23/CIP-23.md
oed: ty

CIPs/CIP-23/CIP-23.md
michaelsena: @oed I think host should be required, in some cases though, host and ID could be the same. This would be the case, for example, if you wanted to verify a root domain. So if this is true, then I think what @stbrody is saying could make sense.

oed: But if you are looking for a root domain you would want to filter by protocol (DNS) rather than the actual host so having the host as map doesn't really make sense in that case? Actually unclear what host would be in the case of DNS 🤔

stbrody: Do we expect each type of supported AKA link to require being defined in a CIP, or are we hoping for a generic format that enables arbitrary links we haven't thought about yet?  If the former, we could require each type of link to have a unique identifier for that link type e.g. "twitter", "github", "dns" that could be the key in the map.  If we support arbitrary links though then it's harder to guarantee that everyone uses the same identifier for the same type of link.  Although in that case the structure of the link documents could be arbitrary also, or use slightly different "host" field values, so maybe we're no worse off?

oed: I think the original intent behind this was that it should support arbitrary new types of entries. *protocol*, *host*, and *id* is fairly generic and you could likely squeeze most things into it 🤷

stbrody (approved):

Issue #69, opened 2020-11-02
Batched Anchor Data Structure


oed:
```
cip: <to be assigned>
title: Batched Anchor Data Structure
author: Joel Thorstensson (@oed)
discussions-to: <URL of the Github issue for this CIP (if this is a PR)>
status: Draft
category: Standards
type: Core
created: 2020-11-02
```

## Simple Summary
This CIP describes the data structure of a blockchain anchor that batches multiple document updates together. 

## Abstract
An anchor service accepts document update anchor request from clients and batches them together in a single merkle tree, which is then anchored into a specific blockchain. The anchor service also sorts all of the requests and creates a bloom filter which describes which documents where updated in this batch, which is also included in the blockchain transaction. This allows external observers to more effectively determine if a document which they are interested in was updated.

## Motivation
There are two main motivations for this CIP. Firstly having an agreed upon spec for the structure of the batched anchor merkle tree makes it easier for all consumers of the ceramic protocol to know what to expect when parsing an *anchor record*. Secondly the sorting of updates and the inclusion of a bloom filter allows indexing services to effectively determine if an update was made which is relevant, and find this update in the tree effectively.

## Specification

First lets define the basic IPLD data structures used.
```ipldsch
type MerkleNode [Link]

type TreeMetaData struct {
  numEntries Int
  bloomType String
  bloomFilter {String:Any}
}
```
Here the *MerkleNode* is simply an array of CIDs, which is used as the main building block for the merkle tree. The *TreeMetaData* contains some metadata about the tree, namely the number of entries and a bloom filter. The `bloomType` is used to describe which specific type of bloom filter is being used and is there for upgradeability purposes.

### Tree structure
The structure of the merkle tree is very straight forward. Starting from the bottom each individual update is paired with another update and a *MerkleNode* is created which has the leftmost update in position `0`, and the other update at position `1`. *MerkleNodes* are then created on each level until we reach the root. The root *MerkleNode* has an additional entry at position `2` which simply points to a *TreeMetaData* object. In the figure below we see a simple tree with four leafs.
![merkletree(1)](https://user-images.githubusercontent.com/3909429/97851426-a83bf900-1cf5-11eb-9d9f-e7b568138850.png)
In this example an *IPLD Path* can be used to reach any of the leafs. For example, to reach *Update 3* we would use the path `<root-cid>/1/0`. Similarly we can always get the *TreeMetaData* using the path `<root-cid>/2`.

[...]

PR #74, opened 2020-11-12
cip: Batched Anchor Data Structure

oed:
PR for CIP-69.
Discussion #69

### DIFF
--- a/CIPs/CIP-69/CIP-69.md
+++ b/CIPs/CIP-69/CIP-69.md
@@ -46,7 +46,7 @@ The merkle tree should be constructed as a balanced tree. See the [implementatio
 ### Leaf sorting
 Before creating the tree the leafs (updates) should be sorted. The sorting should be based on a few different properties of the Ceramic document that is being updated and the update record itself:
 
-1. `topic` - sort by the topic in the document metadata
+1. `collection` - sort by the topic in the document metadata
 2. `schema` - if multiple documents have the same topic, sort by the *schema*
 3. `DID` - if multiple documents have the same topic and schema, extract and sorty by the DID from the `kid` in the signed record
 4. `DocID` - finally sort by the DocID
@@ -55,7 +55,7 @@ Before creating the tree the leafs (updates) should be sorted. The sorting shoul
 ### Bloom filter
 The bloom filter should include the following data for each document that is being updated:
 
-* The `topic`, prepend each tag string with `topic-`
+* The `collection`, prepend each tag string with `collection-`
 
 * The first 5 `tags`, prepend each tag string with `tag-`
 * The `schema`, prepend the schema DocID string with `schema-`
--- a/CIPs/CIP-69/CIP-69.md
+++ b/CIPs/CIP-69/CIP-69.md
@@ -44,23 +44,23 @@ The merkle tree should be constructed as a balanced tree. See the [implementatio
 
 
 ### Leaf sorting
-Before creating the tree the leafs (updates) should be sorted. The sorting should be based on a few different properties of the Ceramic document that is being updated and the update record itself:
+Before creating the tree the leafs (updates) should be sorted. The sorting should be based on a few different properties of the Ceramic document that is being updated:
 
 1. `collection` - sort by the topic in the document metadata
 2. `schema` - if multiple documents have the same topic, sort by the *schema*
-3. `DID` - if multiple documents have the same topic and schema, extract and sorty by the DID from the `kid` in the signed record
+3. `controllers` - if multiple documents have the same topic and schema, sort by the first controller, then subsequent ones
 4. `DocID` - finally sort by the DocID
 
 
 ### Bloom filter
 The bloom filter should include the following data for each document that is being updated:
 
-* The `collection`, prepend each tag string with `collection-`
+* The `collection`, prepend each string with `collection-`
 
 * The first 5 `tags`, prepend each tag string with `tag-`
 * The `schema`, prepend the schema DocID string with `schema-`
-* All DID strings in the `controllers` array, prepend each DID with `did-`
-* The DocID string of the document
+* All DID strings in the `controllers` array, prepend each DID with `controller-`
+* The DocID string of the document, prepend with `docid-`
 
 The bloom filter is created using the javascript [bloom-filter](https://github.com/Callidon/bloom-filters) library. Specifically using the *Classic Bloom Filter*. An example for how to create the filter can be observed below.
 
@@ -84,9 +84,9 @@ Now we can create the *TreeMetaData* object, which can be added into ipfs and th
 
 
 ## Rationale
-This CIP provides the most efficient merkle tree implementation possible in IPLD. The more interesting aspects of this CIP are however the sorting algorithm and the bloom filter implementation. 
+This CIP provides the most efficient merkle tree implementation possible in IPLD DAG-COSE. The more interesting aspects of this CIP are however the sorting algorithm and the bloom filter implementation.
 
-The sorting allows for more efficient lookup in the merkle tree, primarily based on the primary tag of the updated Ceramic document. Further sorting is also provided by the additional properties. If an implementer knows the first tag of a document they can do a binary search though the tree to find the correct update.
+The sorting allows for more efficient lookup in the merkle tree, primarily based on the collection of the updated Ceramic document. Further sorting is also provided by the additional properties. If an implementer knows the collection of a document they can do a binary search though the tree to find the correct update.
 
 The bloom filter allows an observer to efficiently determine if any of the given properties are most likely in the merkle tree. Since the filter itself is stored in a separate IPLD object the filter doesn't add any overhead when verifying a merkle witness for a document update. The bloomfilter implementation that was chosen is currently only implemented in javascript, we failed to find an implementation that is compatible with multiple languages. However, since we introduce a `bloomType` property we can easily determine which implementation was used. This also allows us to change the implementation used in the future.
 
--- a/CIPs/CIP-69/CIP-69.md
+++ b/CIPs/CIP-69/CIP-69.md
@@ -40,7 +40,7 @@ The structure of the merkle tree is very straight forward. Starting from the bot
 ![merkletree(1)](https://user-images.githubusercontent.com/3909429/97851426-a83bf900-1cf5-11eb-9d9f-e7b568138850.png)
 In this example an *IPLD Path* can be used to reach any of the leafs. For example, to reach *Update 3* we would use the path `<root-cid>/1/0`. Similarly we can always get the *TreeMetadata* using the path `<root-cid>/2`.
 
-The merkle tree should be constructed as a balanced tree. See the [implementation][https://github.com/ceramicnetwork/ceramic-anchor-service/blob/4615c814ed98bb17cba8902480d8f22bde5397a1/src/merkle/merkle-tree.ts#L46-L65] for reference.
+The merkle tree should be constructed as a balanced tree. See the [implementation][https://github.com/ceramicnetwork/ceramic-anchor-service/blob/ced4ea9f8c70aa09e6c87e3c1b2de5bbdf505157/src/merkle/merkle-tree.ts#L26-L61] for reference.
 
 
 ### Leaf sorting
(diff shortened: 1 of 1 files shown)


CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+### Bloom filter
+The bloom filter should include the following data for each document that is being updated:
+
+* The `topic`, prepend each tag string with `topic-`
+
+* The first 5 `tags`, prepend each tag string with `tag-`
+* The `schema`, prepend the schema DocID string with `schema-`
+* All DID strings in the `controllers` array, prepend each DID with `did-`
```
stbrody, line 62: might we ever want to include DIDs that aren't the controllers?  I feel like to be safe the prefix should be `controller-did`

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+* The DocID string of the document
```
stbrody, line 63: `docid-` prefix?

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+
+The bloom filter is created using the javascript [bloom-filter](https://github.com/Callidon/bloom-filters) library. Specifically using the *Classic Bloom Filter*. An example for how to create the filter can be observed below.
+
+```js
+const { BloomFilter } = require('bloom-filters')
+const filterEntries = [...] // An array of strings
+const errorRate = 0.04 // 4 % error rate
```
stbrody, line 70: Why .04?  Was that chosen arbitrarily or based on some calculations about the size of the resulting filter?

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Specification
+
+First lets define the basic IPLD data structures used.
+```ipldsch
+type MerkleNode [Link]
+
+type TreeMetaData struct {
+  numEntries Int
+  bloomType String
```
stbrody, line 32: I wonder if we want more than just a string for specifying the bloomType here.  In order to consume the bloom filter users will have to know the arguments used to construct it (either size of the bloom filter and number of hash functions, or target number of items and false positive rate).  To provide that to them we'll either need a public mapping of the `bloomType` string to the arguments used to construct it (so that we could change the `bloomType` string if we ever wanted to change the arguments we're using to construct the bloom filter, even if the underlying bloom filter library we're using doesn't change) or we need to encode those arguments here.  I think putting them here probably makes more sense

stbrody: Actually we have to put it here, because there's no way to know how many entries are in the bloom filter otherwise.  Different anchor merkle trees could wind up with a different number of filter entries, and if we're generating the bloom filter based on the exact known number of entries, then that's going to change the size of the filter.

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+  bloomType: 'npmjs_bloom-filter',
+  bloomFilter: exported
+}
+```
+
+
+## Rationale
+This CIP provides the most efficient merkle tree implementation possible in IPLD. The more interesting aspects of this CIP are however the sorting algorithm and the bloom filter implementation. 
+
+The sorting allows for more efficient lookup in the merkle tree, primarily based on the primary tag of the updated Ceramic document. Further sorting is also provided by the additional properties. If an implementer knows the first tag of a document they can do a binary search though the tree to find the correct update.
```
stbrody, line 89: This sentence is out of date now, still talks about sorting by tag

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Tree structure
+The structure of the merkle tree is very straight forward. Starting from the bottom each individual update is paired with another update and a *MerkleNode* is created which has the leftmost update in position `0`, and the other update at position `1`. *MerkleNodes* are then created on each level until we reach the root. The root *MerkleNode* has an additional entry at position `2` which simply points to a *TreeMetaData* object. In the figure below we see a simple tree with four leafs.
+![merkletree(1)](https://user-images.githubusercontent.com/3909429/97851426-a83bf900-1cf5-11eb-9d9f-e7b568138850.png)
+In this example an *IPLD Path* can be used to reach any of the leafs. For example, to reach *Update 3* we would use the path `<root-cid>/1/0`. Similarly we can always get the *TreeMetaData* using the path `<root-cid>/2`.
+
+The merkle tree should be constructed as a balanced tree. See the [implementation][https://github.com/ceramicnetwork/ceramic-anchor-service/blob/4615c814ed98bb17cba8902480d8f22bde5397a1/src/merkle/merkle-tree.ts#L46-L65] for reference.
```
stbrody, line 43: So reading that implementation, I don't actually think it produces a balanced tree.  I filed https://github.com/ceramicnetwork/ceramic-anchor-service/issues/138 describing the issue.

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+### Leaf sorting
+Before creating the tree the leafs (updates) should be sorted. The sorting should be based on a few different properties of the Ceramic document that is being updated and the update record itself:
+
+1. `topic` - sort by the topic in the document metadata
+2. `schema` - if multiple documents have the same topic, sort by the *schema*
+3. `DID` - if multiple documents have the same topic and schema, extract and sorty by the DID from the `kid` in the signed record
```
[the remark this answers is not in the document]
oed: It's a field in the protected header of the JWS. It contains the DID as well as some query params for the DID version (which should be stripped here)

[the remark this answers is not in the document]
oed: I think we might have to remove this since in the future the signed record might contain multiple signatures, which means multiple `kid`s.

CIPs/CIP-69/CIP-69.md
oed: Kind of arbitrarily. Just took it out of the readme of the `bloom-filters` package.

CIPs/CIP-69/CIP-69.md
oed: Thanks, will have to update here once fixed!

CIPs/CIP-69/CIP-69.md
oed: The library we intend on using will export to a format which includes this information: https://github.com/Callidon/bloom-filters#export-and-import
So the info will be in the `bloomFilter` property.

CIPs/CIP-69/CIP-69.md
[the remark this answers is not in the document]
oed: Updated this to use the controllers array again.

CIPs/CIP-69/CIP-69.md
```diff
@@ -44,23 +44,23 @@ The merkle tree should be constructed as a balanced tree. See the [implementatio
...
 
 ### Leaf sorting
-Before creating the tree the leafs (updates) should be sorted. The sorting should be based on a few different properties of the Ceramic document that is being updated and the update record itself:
+Before creating the tree the leafs (updates) should be sorted. The sorting should be based on a few different properties of the Ceramic document that is being updated:
 
-1. `topic` - sort by the topic in the document metadata
+1. `collection` - sort by the topic in the document metadata
 2. `schema` - if multiple documents have the same topic, sort by the *schema*
-3. `DID` - if multiple documents have the same topic and schema, extract and sorty by the DID from the `kid` in the signed record
+3. `controllers` - if multiple documents have the same topic and schema, sort by the first controller, then subsequent ones
```
stbrody, line 51: For this to work right, we'll have to enforce that the `controllers` array in records is always stored sorted, regardless of the order the user specifies them in.  Should we make a story for this now so we don't forget?  Is there an epic for supporting multiple controllers we could put it in?

CIPs/CIP-69/CIP-69.md
```diff
@@ -84,9 +84,9 @@ Now we can create the *TreeMetaData* object, which can be added into ipfs and th
 
 
 ## Rationale
-This CIP provides the most efficient merkle tree implementation possible in IPLD. The more interesting aspects of this CIP are however the sorting algorithm and the bloom filter implementation. 
+This CIP provides the most efficient merkle tree implementation possible in IPLD DAG-COSE. The more interesting aspects of this CIP are however the sorting algorithm and the bloom filter implementation.
```
stbrody, line 87: I assume DAG-COSE should be DAG-JOSE?

CIPs/CIP-69/CIP-69.md
stbrody: Ah I see.  I guess I had been thinking this was supposed to be an example of how someone might consume the information stored in the `bloomFilter` field of the `TreeMetadata` to reconstitute the bloom filter and use it, but I guess this is more like example code of how **we** might build the bloom filter before persisting it in the TreeMetadata.  In that case, is this example even really relevant/useful to include in this document?

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+There are two main motivations for this CIP. Firstly having an agreed upon spec for the structure of the batched anchor merkle tree makes it easier for all consumers of the ceramic protocol to know what to expect when parsing an *anchor record*. Secondly the sorting of updates and the inclusion of a bloom filter allows indexing services to effectively determine if an update was made which is relevant, and find this update in the tree effectively.
+
+
+## Specification
+
+First lets define the basic IPLD data structures used.
+```ipldsch
+type MerkleNode [Link]
+
+type TreeMetaData struct {
```
stbrody, line 30: I believe 'Metadata' is one word and thus the 'D' should be lowercase

CIPs/CIP-69/CIP-69.md
oed: Thanks, should be DAG-CBOR!

CIPs/CIP-69/CIP-69.md
oed: I guess that's true, unless you already know the order from a previous version of the document. I don't think there is an epic yet, but feel free to create one1

CIPs/CIP-69/CIP-69.md
oed: i don't think it hurts to include?

oed:
@stbrody Pushed some updates here.

CIPs/CIP-69/CIP-69.md
stbrody: https://github.com/ceramicnetwork/js-ceramic/issues/514

CIPs/CIP-69/CIP-69.md
stbrody: Only risk is confusion/distraction from the main focus of how to use/consume this structure.  Probably not a big deal either way though

stbrody (approved):

stbrody:
LGTM!

CIPs/CIP-69/CIP-69.md
```diff
@@ -0,0 +1,108 @@
...
+## Specification
+
+First lets define the basic IPLD data structures used.
+```ipldsch
+type MerkleNode [Link]
+
+type TreeMetadata struct {
+  numEntries Int
+  bloomType String
+  bloomFilter {String:Any}
```
stbrody, line 33: Can't seem to make a mult-line suggestion in github but how about instead of separate top-level `bloomType` and `bloomFilter` fields we have a single top-level `bloomFilter` field whose value is an object with a `type` and a `data` field?  So then we can add more tree metadata in the future and it's not mixed-in with bloom filter specific data

oed: Sure.

CIPs/CIP-69/CIP-69.md
oed: @stbrody can you point me to the updated code snippet so that i can update and merge this?

stbrody: https://github.com/ceramicnetwork/ceramic-anchor-service/blob/ced4ea9f8c70aa09e6c87e3c1b2de5bbdf505157/src/merkle/merkle-tree.ts#L26-L61

PR #78, opened 2020-12-16
Image metadata for basic profile

PaulLeCam:
As discussed with @oed, this would replace the need for a custom IPFS-based solution to support multiple image formats, instead allowing to reference multiple sources directly in the basic profile schema.


CIPs/CIP-19/CIP-19.md
```diff
@@ -74,6 +74,43 @@ The Basic Profile schema defines the format of a document that contains the prop
...
+        },
+        "width": {
+          "$ref": "#/definitions/positiveInteger"
+        },
+        "height": {
+          "$ref": "#/definitions/positiveInteger"
+        },
+        "size": {
+          "$ref": "#/definitions/positiveInteger"
+        }
```
oed, line 96: Is this the file size? Do we need that?

oed (approved):
Seems good. No mimetype?

CIPs/CIP-19/CIP-19.md
PaulLeCam: It's not a required field but could be useful for consumer, for example when loading over mobile networks.

PaulLeCam:
I forgot about the mimeType, thanks for noticing! I just added it.

PR #79, opened 2021-02-12
cip: 3ID DID Method Specification

oed:
This PR introduces a CIP for the 3ID DID Method Specification

### DIFF
--- a/CIPs/CIP-79/cip-79.md
+++ b/CIPs/CIP-79/cip-79.md
@@ -1,8 +1,8 @@
 ---
-cip: <to be assigned>
+cip: 79
 title: 3ID DID Method Specification
 author: Joel Thorstensson (@oed)
-discussions-to: <URL of the Github issue for this CIP (if this is a PR)>
+discussions-to: https://github.com/ceramicnetwork/CIP/issues/80
 status: Draft
 category: Standards
 type: RFC
--- a/CIPs/CIP-79/cip-79.md
+++ b/CIPs/CIP-79/cip-79.md
@@ -21,7 +21,7 @@ A DID that uses this method MUST begin with the following prefix: `did:3`. Per t
 
 ## Method Specific Identifier
 
-There are two versions of 3IDs, both versions use a Ceramic document as a way to update the DID document. 3IDv1 is the most recent version, 3IDv0 is supported for legacy reasons. If you are creating new 3IDs you should always use 3IDv1.
+There are two versions of 3IDs, both versions use a Ceramic document as a way to update the DID document. 3IDv1 is the most recent version, 3IDv0 is supported for legacy reasons. If you are creating new 3IDs you should always use 3IDv1. Both versions are always encoded using multibase. To determine if it's v1 or v0 first convert the multibase string into a byte array, if the first varint is `0x01` we have a 3IDv0, and if it's a `0xce` we have a 3IDv1.
 
 ### 3IDv1
 
@@ -42,7 +42,7 @@ did:3:kjzl6cwe1jw149tlplc4bgnpn1v4uwk9rg9jkvijx0u0zmfa97t69dnqibqa2as
 The method specific identifier for 3IDv0 is a [CID](https://github.com/multiformats/cid) as produced by the [IPLD](https://github.com/ipld/specs) codec [dag-cbor](https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md).
 
 ```
-3idv1 = "did:3:<cid>"
+3idv0 = "did:3:<cid>"
 ```
 
 #### Example
@@ -55,7 +55,7 @@ did:3:bafyreiffkeeq4wq2htejqla2is5ognligi4lvjhwrpqpl2kazjdoecmugi
 
 In this section the CRUD operations for a 3ID DID are defined.
 
-### Create (Register)
+### Create
 
 A `3` DID is created by simply creating a Ceramic [`tile`](https://github.com/ceramicnetwork/CIP/blob/master/CIPs/CIP-8/CIP-8.md) document. The [`tile`](https://github.com/ceramicnetwork/CIP/blob/master/CIPs/CIP-8/CIP-8.md) document type takes a DID as the *controller*, and it's recommended that a `did:key` is used. The *controller* is the DID which is allowed to update the document. The *family* of the document is set to `3id`, and the *deterministic* flag is set to `true`.
 
@@ -85,7 +85,7 @@ const doc = await ceramic.createDocument({
 const didString = `did:3:${doc.id}`
 ```
 
-### Read (Resolve)
+### Read/Verify
 
 Resolving a 3ID is quite straight forward. It is done by taking the DocID from the method specific identifier, looking up the referred document using Ceramic, and converting the content of the Ceramic document into a DID document.
 
@@ -111,7 +111,7 @@ const doc = await ceramic.createDocument({
 })
 ```
 
-#### Converting the Ceramic document to the DID document
+#### Converting the loaded Ceramic document to the DID document
 
 Once we have the Ceramic document loaded, load the latest *AnchorCommit* of the document (the latest commit that was anchored to a blockchain). Then get the content of the document which will look something like this:
 
@@ -133,7 +133,7 @@ To convert this into a DID document first create an empty DID document:
   "publicKey": [],
   "authentication": [],
   "keyAgreement": []
-  }
+}
 ```
 
 Now iterate though the entires in the `publicKeys` object in the Ceramic document and do the following:
@@ -200,25 +200,25 @@ When resolving the DID document [DID Document Metadata](https://w3c.github.io/di
 
 #### Resolving using the `versionId` parameter
 
-When the `versionId` query parameter is given as a DID is resolved it means that we should try to resolve a specific version of the DID document. The resolution process is the same except that the *AnchorCommit* we use for the state of the document should be equal to the DocID + CID from `versionId`. In addition we should construct the *DID Document Metadata* differently.
+When the `versionId` query parameter is given as a DID is resolved it means that we should try to resolve a specific version of the DID document. The resolution process is the same except that the *AnchorCommit* we use to get the content of the document should be equal to the DocID + CID from `versionId`. In addition we should construct the *DID Document Metadata* differently.
 
 ##### DID Document Metadata
 
 * `create` - should be populated using the blockchain timestamp of the first *AnchorCommit*
 * `updated` - should be populated using the blockchain timestamp of the resolved *AnchorCommit*
 * `versionId` - should be equal to the commit CID of the resolved *AnchorCommit*
-* `nextUpdate` - should be populated using the blockchain timestamp of the next *AnchorCommit*
-* `nextVersionId` - should be equal to the commit CID of the next *AnchorCommit*
+* `nextUpdate` - should be populated using the blockchain timestamp of the next *AnchorCommit* (if present)
+* `nextVersionId` - should be equal to the commit CID of the next *AnchorCommit* (if present)
 
 ### Update
 
 The 3ID DID can be updated by changing the content of the Ceramic document corresponding the particular 3ID. Any number of public key can be added or removed from the document content. Note that the *controller* of the Ceramic document can be changed as well. This does not have any effect on the state of the DID document, but changes the DID which is in control of the 3ID document.
 
-### Delete (Deactivate)
+### Deactivate
 
 The 3ID can be deactivated by removing all content in the Ceramic document of the 3ID and replacing it with one property `deactivated` set to `true`.
 
-## Extensability
+## Extensibility
 
 As can be seen in the CRUD section, currently only `secp256k1` and `x25519` public keys are supported. This can be easily extended by using other multicodec encoded keys. The 3ID DID Method could also easily be extended to support other features specified in  [did-core](https://w3c.github.io/did-core/), e.g. service endpoints.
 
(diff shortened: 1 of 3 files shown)


CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+did:3:bafyreiffkeeq4wq2htejqla2is5ognligi4lvjhwrpqpl2kazjdoecmugi
+```
+
+## CRUD Operation Definitions
+
+In this section the CRUD operations for a 3ID DID are defined.
+
+### Create (Register)
+
+A `3` DID is created by simply creating a Ceramic [`tile`](https://github.com/ceramicnetwork/CIP/blob/master/CIPs/CIP-8/CIP-8.md) document. The [`tile`](https://github.com/ceramicnetwork/CIP/blob/master/CIPs/CIP-8/CIP-8.md) document type takes a DID as the *controller*, and it's recommended that a `did:key` is used. The *controller* is the DID which is allowed to update the document. The *family* of the document is set to `3id`, and the *deterministic* flag is set to `true`.
```
stbrody, line 60: I feel like we should say something about what the controller key should be and what it represents and what its relationship is to the keys in the `publicKeys` array in the content body

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+Below you can see an example of how to create a 3ID using the Ceramic javascript api.
+
+```js
+const doc = await ceramic.createDocument({
+  metadata: {
+    controllers: ['did:key:zQ3shQNcackrTByiYaPGso1Nt7b6r1gSMg4XXBmavzvTMqX1h'], // secp256k1 did:key
+    family: '3id'
+  },
+  content: {
+    publicKeys: {
```
stbrody, line 77: Why are we duplicating part of the key between the key and the value?  Why not either just an array of the full public keys, or if we care about being able to quickly look up a specific key it could be a map with the full public key as the key, and something irrelevant like `true` as the value

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Loading 3IDv1 document
+
+To load a 3IDv1 document simply take the DocID from the method specific identifier and load the document from ceramic.
+
+#### Loading 3IDv0 document
+
+To load a 3IDv0 document, first take the CID from the method specific identifier and load the corresponding object from the IPFS network. The resolved object will contain a `publicKey` property which has an array of public key objects (see [Appendix A](#appendix-a)).
+
+Now find the key with an id that ends in `#signingKey` and convert it to a multicodec encoded key (in compressed format), then use it as a `did:key` and look up the ceramic document which has this key as the *controller* and *family* set to `3id`, using the *deterministic* flag. Below you can see the code to do this with the javascript Ceramic api.
```
stbrody, line 100: wait, so there are actually two ceramic 3id documents that need to be looked up?  So you take the did specifier, get the docId from it, look it up on ceramic, then find the `signingKey` (how do you know which of the keys is the signing key?), then you look up *another* 3id doc based on the signing key?  Now the rest of the document says "Once we have the Ceramic document loaded" but now we have two ceramic documents in play, right?  Which document is being referred to below?

I feel like I must not be understanding this section right

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+    controllers: [didKey], // secp256k1 did:key
+    family: '3id'
+  },
+  deterministic: true
+})
+```
+
+#### Converting the Ceramic document to the DID document
+
+Once we have the Ceramic document loaded, load the latest *AnchorCommit* of the document (the latest commit that was anchored to a blockchain). Then get the content of the document which will look something like this:
```
stbrody, line 116: Here we say just to load the latest `AnchorCommit` but below we reference both the earliest and latest `AnchorCommit`

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+
+When resolving the DID document [DID Document Metadata](https://w3c.github.io/did-core/#did-document-metadata) should be provided. When resolving a 3ID we should populate the following fields:
+
+* `create` - should be populated using the blockchain timestamp of the first *AnchorCommit*
+* `updated` - should be populated using the blockchain timestamp of the most recent *AnchorCommit*
+* `versionId` - should be equal to the commit CID of the most recent *AnchorCommit*
+
+#### Resolving using the `versionId` parameter
+
+When the `versionId` query parameter is given as a DID is resolved it means that we should try to resolve a specific version of the DID document. The resolution process is the same except that the *AnchorCommit* we use for the state of the document should be equal to the DocID + CID from `versionId`. In addition we should construct the *DID Document Metadata* differently.
```
stbrody, line 203: We also need to take the **content** from that version of the document, not just the `AnchorCommit` metadata

stbrody: Not sure who exactly the intended audience here is.  For us it's probably obvious that we want to actually load the document contents at that specific version, but if this is for someone from w3c that may not be as obvious.  If someone who wasn't us tried to implement this off of the spec, I could see them loading the current version of the document, but then just iterating the log and taking the `AnchorCommit` metadata as you specify here, but leaving the `content` from the current version.  So that's why I'm being extra pedantic and pointing this out.  If you think it's not a concern for how this doc will be used, then that's fine and feel free to leave it as-is!

stbrody (commented):
Some questions inline

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+```sh
+did:3:kjzl6cwe1jw149tlplc4bgnpn1v4uwk9rg9jkvijx0u0zmfa97t69dnqibqa2as
+```
+
+### 3IDv0
+
+The method specific identifier for 3IDv0 is a [CID](https://github.com/multiformats/cid) as produced by the [IPLD](https://github.com/ipld/specs) codec [dag-cbor](https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md).
+
+```
+3idv1 = "did:3:<cid>"
```
stbrody, line 45:
```suggestion
3idv0 = "did:3:<cid>"
```

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+
+## DID Method Name
+
+The name string that shall identify this DID method is: `3`.
+
+A DID that uses this method MUST begin with the following prefix: `did:3`. Per the [DID specification](https://w3c.github.io/did-core/), this string MUST be in lowercase. The remainder of the DID, after the prefix, is specified below.
+
+## Method Specific Identifier
+
+There are two versions of 3IDs, both versions use a Ceramic document as a way to update the DID document. 3IDv1 is the most recent version, 3IDv0 is supported for legacy reasons. If you are creating new 3IDs you should always use 3IDv1.
```
stbrody, line 24: Do we need to explain how to know if a 3ID is v0 or v1?  I assume we try to parse the identifier as a DocID and if that fails then we parse it as a CID and assume we have a v0 id?

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Update
+
+The 3ID DID can be updated by changing the content of the Ceramic document corresponding the particular 3ID. Any number of public key can be added or removed from the document content. Note that the *controller* of the Ceramic document can be changed as well. This does not have any effect on the state of the DID document, but changes the DID which is in control of the 3ID document.
+
+### Delete (Deactivate)
+
+The 3ID can be deactivated by removing all content in the Ceramic document of the 3ID and replacing it with one property `deactivated` set to `true`.
+
+## Extensability
```
stbrody, line 221:
```suggestion
## Extensibility
```

CIPs/CIP-79/cip-79.md
oed: What do you mean with "what the controller key should be"? It doesn't need to have any relationship to the keys in the `publicKeys` array.

CIPs/CIP-79/cip-79.md
oed: This basically allows someone to have a specific name for a key, even if that's not recommended.

CIPs/CIP-79/cip-79.md
oed: No 3IDv0 loads the genesis object from IPFS:
> To load a 3IDv0 document, first take the CID from the method specific identifier and load the corresponding object from the IPFS network.

For 3IDv1 you just load the docId in the method specific identifier.

CIPs/CIP-79/cip-79.md
stbrody: Right, I'm thinking we should probably say that explicitly, and clarify what the relationship between the controller and the 3id is.  So basically the `publicKeys` are allowed to sign and encrypt messages with this 3id, but cannot change the set of keys that are allowed to do so.  The `controller` is the super-user who can modify the set of keys that can authenticate as the 3id (and therefore could always just grant itself access to authenticate as the 3id if it wanted)

CIPs/CIP-79/cip-79.md
oed: Well yeah. Once you have loaded the latest `AnchorCommit` you should be able to easily get the data from the earliest one?

CIPs/CIP-79/cip-79.md
oed: > The resolution process is the same except that the *AnchorCommit* we use for the **state** of the document should be equal to the DocID + CID from `versionId`.

Imo, **state** implies the *content*. 

The audience of this is really someone that will be implementing the 3ID did method.

CIPs/CIP-79/cip-79.md
oed: Good point. I'll add an explainer for this 👍

CIPs/CIP-79/cip-79.md
oed: Also would be a pain to change it at this point.

stbrody: oh interesting.  I guess the spec does use the word "should" - not "must" - when describing the structure of this object.  And I suppose you're not mentioning that in the spec because you don't want to encourage it?  But if we don't want to encourage it, then why are we allowing it?  Shouldn't we pick a position and go either one way or the other?

CIPs/CIP-79/cip-79.md
oed: Clarified this to be more clear about *content*.

CIPs/CIP-79/cip-79.md
stbrody: > Also would be a pain to change it at this point.

Yep, totally fair.  Just wanted to understand if this is here because we actually want it this way, or just because of historical decisions we no longer stand by.

CIPs/CIP-79/cip-79.md
stbrody: Ugh I must have somehow forgotten I was reading about V0 midway through reading this section.  This all makes sense now, sorry for the confusion

CIPs/CIP-79/cip-79.md
stbrody: its just a weird asymmetry.  It's kind of confusing that we mention loading the `AnchorCommit` here at all when we don't actually do anything with it until several steps later.  Maybe just remove the reference from this section entirely and just make sure to define the term `AnchorCommit` later when you actually need it for the first time?

CIPs/CIP-79/cip-79.md
stbrody: Okay, if you're not worried about it then I won't either

CIPs/CIP-79/cip-79.md
oed: I don't feel strongly that we must pick a position.

CIPs/CIP-79/cip-79.md
oed: > The *controller* is the DID which is allowed to update the document.

This not clear enough?

Added a sentence on the public keys.

CIPs/CIP-79/cip-79.md
oed: We do do something with it since we use the content from the latest anchorCommit. Clarified this a bit now.

stbrody (approved):
LGTM!  Nice work!

CIPs/CIP-79/cip-79.md
```diff
@@ -0,0 +1,282 @@
...
+
+Resolving a 3ID is quite straight forward. It is done by taking the DocID from the method specific identifier, looking up the referred document using Ceramic, and converting the content of the Ceramic document into a DID document.
+
+#### Loading 3IDv1 document
+
+To load a 3IDv1 document simply take the DocID from the method specific identifier and load the document from ceramic.
+
+#### Loading 3IDv0 document
+
+To load a 3IDv0 document, first take the CID from the method specific identifier and load the corresponding object from the IPFS network. The resolved object will contain a `publicKey` property which has an array of public key objects (see [Appendix A](#appendix-a)).
```
zachferland, line 98: somewhat covered by `3IDv0 genesis` section below, but maybe note what happens if no ceramic doc is resolved in the steps here

zachferland (approved):
Looks good 👍

CIPs/CIP-79/cip-79.md
oed: Well, the genesis commit will be resolved regardless since it's deterministically generated. Will see if I can improve it a bit!

PR #85, opened 2021-02-16
cip: AppendCollection

PaulLeCam:


PaulLeCam:
Merged in https://github.com/ceramicnetwork/CIP/pull/89

CIPs/CIP-85/cip-85.md
```diff
@@ -0,0 +1,193 @@
...
+#### Other insertions
+
+1. Load the `AppendCollection` document.
+1. Load the `CollectionSlice` document from the `last` field of the `AppendCollection`.
+1. Check the length of the `contents` array of the `CollectionSlice`:
+
+- If it is lower than the `sliceSize` value of the `AppendCollection`, add the item to the `contents` array.
+- If it is equal to the `sliceSize` value:
+  1. Create a new `CollectionSlice` document with the `contents` array containing the item to insert and the `prev` value containing the DocID of the previous slice.
+  1. Update the previous slice to set the `next` value to the DocID of the newly created slice.
```
stbrody, line 123: I'm a bit concerned about the lack of atomicity between this step and the next one.  I worry that conflicting updates, or a crash in the middle of this process, could leave the `next` pointer from the `CollectionSlice` out of sync with the `last` pointer from the `AppendCollection`.

I'm not totally sure what to do about it though.  The `next` pointer is important to have within the `CollectionSlice` so that scanning forward works efficiently.

I wonder, could we take advantage of deterministic documents to make the `CollectionSlice` docIds knowable deterministically from the `AppendCollection` docId and the slice index?

@oed @PaulLeCam

stbrody: Deterministic slice IDs would also have the nice advantage of allowing binary search and other types of scans over the collection other than strict front-to-back or back-to-front

PaulLeCam: Yes good point about the atomicity, my expectation was that these 2 calls would be done in parallel and if either fails there's a retry process, but obviously that's up to the implementation logic and that doesn't really solve the problem of being atomic.

Deterministic slice IDs is a great idea! That would prevent the need for a `next` reference so adding a new slice would only need to update the `AppendCollection`, and that could speedup loading slices as well as this could be done in parallel rather than sequentially.
How do you see it working in the schema please? Just thinking about it a bit I think the deterministic slice should be something like `{ collection: '<collection docID>', sliceNumber: number, contents: [] }` and then adding the content and `prev` reference when updating, do you think that works?

stbrody: Yeah, that looks great for how the genesis commit is generated.

I actually think with this approach you don't need the `prev` reference in `CollectionSlice` since you can always get the previous slice docID by subtracting 1 from the current sliceNumber.  I also think we can then remove the `first` and `last` reference from `AppendCollection`, and just use the `slicesCount` to find the current "last" slice.

PaulLeCam: Yes good point! I updated the CIP in https://github.com/ceramicnetwork/CIP/pull/91 if you can have a look please?

Issue #88, opened 2021-03-01
CIP: Ceramic namespace in JSON schemas


PaulLeCam:
---
cip: <to be assigned>
title: Ceramic namespace in JSON schemas
author: Paul Le Cam (@PaulLeCam)
status: Draft
category: Standards
type: RFC
created: 2021-03-01
---

## Simple Summary

Define a standard extension point for Ceramic-specific metadata in a JSON schema.

## Abstract

This CIP defines a reserved namespace for Ceramic-specific metadata in a JSON schema, along with a reference table for standard uses of this namespace.

## Motivation

As commented in https://github.com/ceramicnetwork/CIP/issues/82#issuecomment-787449788 the `$id` cannot be used to define Ceramic-specific extensions as intended in [CIP-82](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-82/CIP-82.md), so creating a custom namespace for Ceramic-specific metadata should be a safer option to enable further extensions.

### Namespace

A JSON schema property can contain a `$ceramic` field, that must be an object with a unique `type` defined in the following reference table, for example:

```js
{
  type: 'string',
  maxLength: 150,
  $ceramic: {
    type: 'tile',
    schema: '<schema docID or commitID>' ,
  },
}
```

### Reference table 

| Type | CIP                                                                                             | Status     |
| ----- | ------------------------------------------------------------ | --------- |
| `tile`  | [DocID json-schema definition](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-82/CIP-82.md)   | Draft     |

## Rationale

Using the `$ceramic` property should be consistent with other `$`-prefixed metadata properties in JSON schemas, avoiding possible conflicts with other property names.

A unique `type`, along with possible type-specific additional properties, should make it easy to add custom extensions and build tools (simple checks for existence of `$ceramic` property and type-specific logic, TypeScript interfaces and inference, etc.).

Finally, providing a reference table in this CIP should allow for easy discovery and avoid conflicts between extensions.

## Backwards Compatibility

[CIP-82](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-82/CIP-82.md) and [CIP-85](https://github.com/ceramicnetwork/CIP/pull/85) will get updated to this new format.

## Implementation

None yet.

## Security Considerations

None I'm aware of.

## Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

PR #89, opened 2021-03-01
Update CIP-82 to use CIP-88 namespace

PaulLeCam:
CIP-88 issue: https://github.com/ceramicnetwork/CIP/issues/88


oed (commented):
Can you include CIP-88 as well? I guess CIP-82 should state that it requires 88?

PaulLeCam:
Yes good idea, I added the requirement to the header.

oed:
But do you mind also adding CIP-88 as part of this PR? Would be good if it's finable in the INDEX.md if it's a requirement here.

PaulLeCam:
I could but I thought the flow would be first discussing/merging CIP 88 and then updating 82 and 85, no? Basically I'd like to make sure possible comments about 88 are addressed before making further changes to other CIPs.

I don't mind adding 88 in this PR if it makes things simpler, just want to make sure it doesn't slow the process down.

oed:
I think it's fine to merge drafts before much discussion. Drafts can always easily be updated with a new PR if needed.

oed (approved):

PR #93, opened 2021-03-05
Update CIP-69 to match what was actually implemented

stbrody:


oed (approved):
Maybe change status to "Final"?

PR #98, opened 2021-03-22
switched the labels for x25519 and secp256k1

bshambaugh:
I changed the labeling for secp256k1 and x25519 to be in line with https://w3c-ccg.github.io/did-method-key/ . 

I used code https://gist.github.com/bshambaugh/e3730a24cf1f87813c32f7e7f41a1179 . I noticed that x25519 was only 31 bytes rather than 32.

### DIFF
--- a/CIPs/CIP-79/CIP-79.md
+++ b/CIPs/CIP-79/CIP-79.md
@@ -75,8 +75,8 @@ const doc = await ceramic.createDocument({
   },
   content: {
     publicKeys: {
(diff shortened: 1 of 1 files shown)


oed (approved):
Thanks!

oed:
Note that there are some 3ID test vectors available here: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/3id-did-resolver/src/__tests__/vectors.json

Issue #87, opened 2021-02-28
cip: Conviction Voting Template


oed:
```
cip: <to be assigned>
title: Conviction Voting Template
author: Joel Thorstensson (@oed)
discussions-to: <URL of the Github issue for this CIP (if this is a PR)>
status: Idea
category: Standards
type: RFC
created: 2021-02-28
```

> This is still a work in progress. Will add more context to these schemas!

## Simple Summary
This CIP describes a template that can be used to create off-chain conviction voting dapps.

## Abstract

Abstract goes here.

## Motivation

Fix me :)

### Definition content

**Deployment:** Deploy your own.
```json
{
  "name": "Conviction Voting Dapp",
  "description": "Voting data for my dapp",
  "schema": "<convictions-schema-docid>",
  "config": {
    "context": "<caip19-asset-id>"
  }
}
```

### Proposal Schema
**Deployment:** `<proposal-schema-docid>`

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "title": "Convictions",
  "properties": {
    "context": {
      "type": "string"
    },
    "title": {
      "type": "string"
    },
    "currency": {
      "type": "string"
    },
    "amount": {
      "type": "number"
    },
    "beneficiary": {
      "type": "string"
    },
    "description": {
      "type": "string"
    },
    "url": {
      "type": "string"
    }
  },
  "additionalProperties": false,
  "required": [
    "context",
    "title",
    "currency",
    "amount",
    "beneficiary",
    "url"
  ]
}
```

### Convictions Schema
**Deployment:** `<convictions-schema-docid>`
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "title": "Convictions",
  "properties": {
    "context": {
      "type": "string"
    },
    "supply": {
      "type": "number"
    },
    "proposals": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/proposals"
      }
    },
    "convictions": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/convictions"
      }
    }
  },
  "additionalProperties": false,
  "required": [
    "context",
    "proposals",
    "convictions"
  ],
  "definitions": {
    "proposals": {
      "type": "string",
      "$id": "ceramic://doc",
      "$ceramicSchema": "<proposal-schema-docid>",
      "maxLength": 150
    },
    "convictions": {
      "type": "object",
      "properties": {
        "proposal": {
[...]

PR #104, opened 2021-06-14
Update Ceramic metadata field to use $comment field

PaulLeCam:
- Uses `$comment` instead of `$ceramic` field in JSON schemas
- Replaces mentions of document/DocID by stream/StreamID
- Uses `tags` metadata for deterministic slice access for append-collection

### DIFF
--- a/CIPs/CIP-82/CIP-82.md
+++ b/CIPs/CIP-82/CIP-82.md
@@ -1,6 +1,6 @@
 ---
 cip: 82
-title: DocID json-schema definition
+title: StreamID json-schema definition
 author: Paul Le Cam (@PaulLeCam)
 discussions-to: https://github.com/ceramicnetwork/CIP/issues/82
 status: Draft
@@ -13,15 +13,15 @@ requires: [CIP-88](https://github.com/ceramicnetwork/CIP/issues/88)
 
 ## Simple Summary
 
-Provide a static way to define a string in a JSON schema represents a Ceramic DocID, optionally with static references to the schema(s) that must be used by the referenced document.
+Provide a static way to define a string in a JSON schema represents a Ceramic StreamID, optionally with static references to the schema(s) that must be used by the referenced stream.
 
 ## Abstract
 
-This CIP defines a standard way to add a reference to an existing Ceramic document in a JSON schema and references to existing JSON schemas, so it is possible to access this information about Ceramic documents at build time rather than only at runtime on created documents.
+This CIP defines a standard way to add a reference to an existing Ceramic stream in a JSON schema and references to existing JSON schemas, so it is possible to access this information about Ceramic streams at build time rather than only at runtime on created streams.
 
 ## Motivation
 
-It is sometimes necessary to reference Ceramic documents from other documents, such as a list containing the docIDs of individual Ceramic documents.
+It is sometimes necessary to reference Ceramic streams from other streams, such as a list containing the streamIDs of individual Ceramic streams.
 Currently, we are sometimes using definitions that can be referenced in a schema using `"$ref": "#/definitions/CeramicDocId"` for example, but this has not be defined as a standard.
 Using local definitions to a schema also has the downside of providing no guaranty of being unique or having the definition matching any standard.
 
@@ -57,10 +57,7 @@ Using this CIP, a `NotesList` schema could explicitly reference a `Note` schema 
     NoteDocID: {
       type: 'string',
       maxLength: 150,
-      $ceramic: {
-        type: 'tile',
-        schema: '<Note schema docID>',
-      },
+      $comment: 'ceramic:tile:<Note schema streamID>',
     },
   },
 }
@@ -70,25 +67,23 @@ This way, by loading the `Notes` schema, it is possible by a tool/library to dis
 
 ## Specification
 
-References to Ceramic schema should use a `string` with the `$ceramic` field using the `tile` type, and optionally with a `schema` property containing either a string or array of strings containing the DocID (implicit reference to latest version) or CommitID (specific version) of the supported schema(s):
+References to Ceramic schema should use a `string` with the `$comment` field using the `ceramic:tile` type, and optionally with a schema string of the StreamID (implicit reference to latest version) or CommitID (specific version) of the supported schema(s).
+Multiple schemas can be provided, using the `|` character as separator.
 
 ```js
 {
   type: 'string',
   maxLength: 150,
-  $ceramic: {
-    type: 'tile',
-    schema: '<Note schema docID>',
-  },
+  $comment: 'ceramic:tile:<Note schema streamID>',
 }
 ```
 
 ## Rationale
 
-This CIP uses the `$ceramic` namespace defined in [CIP-88](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-88/CIP-88.md).
+This CIP uses the `$comment` field as defined in [CIP-88](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-88/CIP-88.md).
 
 This spec allows to either define a single schema (using a string) or multiple ones (array of strings).
-The use case would be to support different schemas for a single reference, for example a "media" schema could reference an "image" schema, but also the "audio" and "video" ones as acceptable document schemas: `schema: ['<image schema docID>', '<audio schema docID>', '<video schema docID>']`.
+The use case would be to support different schemas for a single reference, for example a "media" schema could reference an "image" schema, but also the "audio" and "video" ones as acceptable document schemas: `$comment: 'ceramic:tile:<image schema docID>|<audio schema docID>|<video schema docID>'`.
 
 ## Backwards Compatibility
 
--- a/CIPs/CIP-82/CIP-82.md
+++ b/CIPs/CIP-82/CIP-82.md
@@ -57,7 +57,7 @@ Using this CIP, a `NotesList` schema could explicitly reference a `Note` schema 
     NoteDocID: {
       type: 'string',
       maxLength: 150,
-      $comment: 'ceramic:tile:<Note schema streamID>',
+      $comment: 'ceramic:doc:<Note schema streamID>',
     },
   },
 }
@@ -67,14 +67,14 @@ This way, by loading the `Notes` schema, it is possible by a tool/library to dis
 
 ## Specification
 
-References to Ceramic schema should use a `string` with the `$comment` field using the `ceramic:tile` type, and optionally with a schema string of the StreamID (implicit reference to latest version) or CommitID (specific version) of the supported schema(s).
+References to Ceramic schema should use a `string` with the `$comment` field using the `ceramic:doc` type, and optionally with a schema string of the StreamID (implicit reference to latest version) or CommitID (specific version) of the supported schema(s).
 Multiple schemas can be provided, using the `|` character as separator.
 
 ```js
 {
   type: 'string',
   maxLength: 150,
-  $comment: 'ceramic:tile:<Note schema streamID>',
+  $comment: 'ceramic:doc:<Note schema streamID>',
 }
 ```
 
@@ -83,7 +83,7 @@ Multiple schemas can be provided, using the `|` character as separator.
 This CIP uses the `$comment` field as defined in [CIP-88](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-88/CIP-88.md).
 
 This spec allows to either define a single schema (using a string) or multiple ones (array of strings).
-The use case would be to support different schemas for a single reference, for example a "media" schema could reference an "image" schema, but also the "audio" and "video" ones as acceptable document schemas: `$comment: 'ceramic:tile:<image schema docID>|<audio schema docID>|<video schema docID>'`.
+The use case would be to support different schemas for a single reference, for example a "media" schema could reference an "image" schema, but also the "audio" and "video" ones as acceptable document schemas: `$comment: 'ceramic:doc:<image schema docID>|<audio schema docID>|<video schema docID>'`.
 
 ## Backwards Compatibility
 
--- a/CIPs/CIP-82/CIP-82.md
+++ b/CIPs/CIP-82/CIP-82.md
@@ -7,7 +7,7 @@ status: Draft
 category: Standards
 type: RFC
 created: 2021-02-15
-edited: 2021-03-01
+edited: 2021-07-02
 requires: [CIP-88](https://github.com/ceramicnetwork/CIP/issues/88)
 ---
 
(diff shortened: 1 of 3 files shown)


CIPs/CIP-82/CIP-82.md
```diff
@@ -70,25 +67,23 @@ This way, by loading the `Notes` schema, it is possible by a tool/library to dis
...
 
 ```js
 {
   type: 'string',
   maxLength: 150,
-  $ceramic: {
-    type: 'tile',
-    schema: '<Note schema docID>',
-  },
+  $comment: 'ceramic:tile:<Note schema streamID>',
```
stbrody, line 77: A part of me wonders if we'd be better off using the streamtype code rather than name (so 0 for tile, 1 for caip10link, etc). I'm worried that once we have a large number of stream types that all represent json documents just with different log structures, coming up with meaningful human readable names for them all is going to get tricky...

CIPs/CIP-85/CIP-85.md
```diff
@@ -17,46 +17,50 @@ Provide schemas and associated tools to help interact with large lists in Cerami
...
-The schemas use document references defined in [CIP-82](https://github.com/ceramicnetwork/CIP/issues/82) to identify relations.
+The schemas use stream references defined in [CIP-82](https://github.com/ceramicnetwork/CIP/issues/82) to identify relations.
 
 #### Cursors
 
-A cursor is a unique identifier for an item in the collection, made of the tuple (CollectionSlice DocID, item index).
+A cursor is a unique identifier for an item in the collection, made of the tuple (CollectionSlice StreamID, item index).
 A collection slice should contain a maximum of 256 items, so the index of an item in the slice can be represented in a single byte.
 
+#### Slice tag
```
stbrody, line 41: Hmm... I wonder if it makes sense to use `tags` in this way, or if we should wait for https://github.com/ceramicnetwork/js-ceramic/issues/1342 and use a more semantically meaningful object structure in the metadata for this

CIPs/CIP-85/CIP-85.md
```diff
@@ -101,37 +105,33 @@ The `CollectionSlice` schema must be an `object` with a `$ceramic` field from [C
 
 #### Deterministic slice access
 
-Accessing slices in the collection relies on Ceramic's ability to load documents deterministically based on their genesis contents:
-
-- `collection`: the collection's `DocID` string
-- `sliceIndex`: the slice's index, an integer between `0` (first slice) and `slicesCount - 1` (last slice)
-- `contents`: empty array
+Accessing slices in the collection relies on Ceramic's ability to load streams deterministically based on their metadata by using `{ deterministic: true, tags: ['<slice tag>'] }`, where `slice tag` is a string of `<collection StreamID string>:<slice index>`.
```
stbrody, line 108: Are there any other tags we might want to include for the purpose of indexing services?  We can't change the tags we use later without breaking deterministic access.

For instance, I could imagine a tag just for the streamId of the parent collection, so indexers could find all slices for a given collection.

stbrody: I guess if we want to add tags later without breaking deterministic access we could do so they would just have to be added in the first update, not in the genesis commit

CIPs/CIP-88/CIP-88.md
```diff
@@ -18,22 +18,22 @@ This CIP defines a reserved namespace for Ceramic-specific metadata in a JSON sc
 
 ## Motivation
 
-As commented in https://github.com/ceramicnetwork/CIP/issues/82#issuecomment-787449788 the `$id` cannot be used to define Ceramic-specific extensions as intended in [CIP-82](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-82/CIP-82.md), so creating a custom namespace for Ceramic-specific metadata should be a safer option to enable further extensions.
+As commented in https://github.com/ceramicnetwork/CIP/issues/82#issuecomment-787449788 the `$id` cannot be used to define Ceramic-specific extensions as intended in [CIP-82](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-82/CIP-82.md).
+
+A previous version of this CIP relied on creating a custom namespace for Ceramic-specific metadata using the non-standard `$ceramic` key, but it was not compatible with the "strict mode" of the JSON validation library used in Ceramic (AJV).
+Instead, this CIP would rely on the `$comment` field that is supported by AJV's strict mode.
```
stbrody, line 24:
```suggestion
Instead, this CIP relies on the `$comment` field that is supported by AJV's strict mode.
```

stbrody (commented):
My main comments are around thinking about how we use and think about the metadata

CIPs/CIP-82/CIP-82.md
PaulLeCam: Yes I agree there might be a better namespace than `tile` but I'm not sure stream types are so relevant here, because the purpose is to interact with other JSON documents anyways?
At this level I don't think it matters much if the JSON doc is backed by a tile, CRDTs or any other stream type, but rather than it can be resolved by the Ceramic client, no?

CIPs/CIP-85/CIP-85.md
PaulLeCam: Sure, we should use whatever makes most sense for Ceramic, I wasn't aware there would be a new field in the metadata for this.
Might be good to clarify the relations and differences between `family`, `tags` and this new field i guess, as I thought `family` was the field meant for indexing (at least in the context of IDX).

CIPs/CIP-85/CIP-85.md
PaulLeCam: Yes we need a first update in any case to add the contents so it shouldn't be a problem to also edit the metadata. I agree having a tag with the collection ID could be useful.

CIPs/CIP-85/CIP-85.md
stbrody: I think the idea is for both family and tags to be used by indexers

CIPs/CIP-85/CIP-85.md
stbrody: will resolve this comment thread as it doesn't seem like this is actually a blocker since it can be changed later

CIPs/CIP-82/CIP-82.md
stbrody: ah right, good point.  In that case, is there any reason to include the streamtype name at all?  Could it just be `ceramic:<streamID>`?  The StreamID already has the streamtype ID number embedded within it.

PaulLeCam: `tile` here was meant as a namespace, as we have `ceramic:appendCollection:<slice schema ID>` and `ceramic:collectionSlice` that have different meaning.
Basically `ceramic:tile` should just mean "this string points to a Ceramic JSON document" and `ceramic:tile:<schema ID>` means a reference to a JSON doc having a specific schema.
Maybe we can just use `ceramic:doc` or `ceramic:ref` instead if it is more explicit?

stbrody: What's gained by this namespacing?  Is it just for human readability? Or does the user or code many any decisions based on the namespace?

PaulLeCam: Purely tooling at the moment for docset and GraphQL support purposes. I don't think human readability matters too much, at least it's not meant for end-users, but might be simpler for debugging.

stbrody: okay, so there *is* code that is going to make decisions based on this namespace?  Can you clarify where/how?

PaulLeCam: Sure, you can see the tile reference being used with docsets to extract schema references here: https://github.com/ceramicstudio/js-idx/blob/feat/append-collection/packages/tools/src/docset.ts#L46-L56
Similar here for an append collection to extract the slice schema: https://github.com/ceramicstudio/js-idx/blob/feat/append-collection/packages/append-collection/src/append-collection.ts#L39-L47

stbrody: Okay, so it seems like it's mostly for sanity check input validation.  Nothing intrinsically that couldn't be done just with `ceramic:<streamid>`, but just adds some extra validation that the right types of streams are being used in the right places.

Something like `ceramic:doc` seems like it would make sense to me, but it kind of depends on some decisions we haven't really made yet on what the relationships will be between different streamtypes that implement the same user-facing APIs (e.g. different stream types that all represent a json document, but with different log structures, conflict resolution mechanisms, history retention, etc).

stbrody: My current vote is for `ceramic:doc:<streamID>`, with the understanding that `ceramic:doc` refers to a streamtype that is one of the streamtypes within the group of streams that expose a JSON document API.

How client/application code will know what is the set of streamtypes that map to that type of API is still an open question however

CIPs/CIP-85/CIP-85.md
stbrody: my vote would be to do this with the new metadata object from https://github.com/ceramicnetwork/js-ceramic/issues/1342 so the field names could have more semantic meaning, but I don't feel super strongly and could be convinced it makes sense to just do this with tags

stbrody:
I've laid out my thoughts as best as I can, but there are still 2 open comment threads which need some final decisions made.  I just made a last comment on each thread summarizing my current thinking.  Now I think we need @oed to weigh in.

CIPs/CIP-82/CIP-82.md
PaulLeCam: 👍 I replaced `tile` by `doc`

PaulLeCam:
@oed @stbrody do you think we could merge this as-is for now please? We can always update the specs based on the implementation changes.

CIPs/CIP-82/CIP-82.md
oed: I don't really like that we invent a new format of encoding streamids here.

I propose that we use the standard url for streamids with an additional query param:
- `ceramic://<streamid>?json-schema=doc`
- `ceramic://<streamid>?json-schema=appendCollection`

oed:
@PaulLeCam made a proposal for how to use our existing url encoding for streamids

CIPs/CIP-85/CIP-85.md
oed: `family` is the primary key used for indexing. I agree with @stbrody's point unless we actually want the indexer to more easily find it using the `tag`.

CIPs/CIP-82/CIP-82.md
PaulLeCam: These specs are about metadata in the schemas themselves, not about the contents.
In `ceramic:doc:<id>` the `id` is the stream ID of the schema the referenced doc should use, not the contents doc.

These strings don't even necessarily represent stream IDs, for example `ceramic:appendCollection` is just a way for the tools to detect a schema represents an AppendCollection.

oed: I get that, but I don't see why we need to introduce another representation. As far as I can tell we always include the StreamID. Any problem with using the formatting I suggested above so that we don't introduce a new representation?

CIPs/CIP-82/CIP-82.md
```diff
@@ -70,25 +67,23 @@ This way, by loading the `Notes` schema, it is possible by a tool/library to dis
...
+  $comment: 'ceramic:doc:<Note schema streamID>',
 }
 ```
 
 ## Rationale
 
-This CIP uses the `$ceramic` namespace defined in [CIP-88](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-88/CIP-88.md).
+This CIP uses the `$comment` field as defined in [CIP-88](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-88/CIP-88.md).
 
 This spec allows to either define a single schema (using a string) or multiple ones (array of strings).
```
oed, line 85: `array of strings` seems incorrect here.

CIPs/CIP-85/CIP-85.md
```diff
@@ -17,46 +17,50 @@ Provide schemas and associated tools to help interact with large lists in Cerami
...
 
-This CIP presents a way to store a possibly large list of items split into multiple documents based on two complementary schemas: `AppendCollection` and `CollectionSlice`.
-This allows to create a virtually infinite list of items such as a feed by leveraging multiple Ceramic documents rather than a single one.
+This CIP presents a way to store a possibly large list of items split into multiple streams based on two complementary schemas: `AppendCollection` and `CollectionSlice`.
+This allows to create a virtually infinite list of items such as a feed by leveraging multiple Ceramic streams rather than a single one.
 
 ## Motivation
 
-Storing a list of items in a single Ceramic document can make the document grow large in size, possibly exceeding Ceramic's internal limits.
+Storing a list of items in a single Ceramic stream can make the stream grow large in size, possibly exceeding Ceramic's internal limits.
```
oed, line 25: This is true for Tiles today, but it not really true for future stream implementations. I would add wording about that this is a specific problem for Tiles.

CIPs/CIP-85/CIP-85.md
```diff
@@ -69,16 +73,16 @@ The `AppendCollection` schema must be an `object` with a `$ceramic` field from [
...
-- `collection`: the DocID of the collection the slice is part of
+- `collection`: the StreamID of the collection the slice is part of
 - `sliceIndex`: index of the slice in the collection, between `0` and the collection's `slicesCount` minus `1`
 - `contents`: array with a `maxItems` value matching the `AppendCollection` `sliceMaxItems` property and defining the items schemas, that must include `{ type: 'null' }` in order to support removals
 
 ```js
 {
   $schema: 'http://json-schema.org/draft-07/schema#',
-  $ceramic: { type: 'collectionSlice' },
+  $comment: 'ceramic:collectionSlice',
```
oed, line 85: Why is this property needed here? I don't see any rationale for it. The fact that the schema is referenced from the append collection should be enough context for us to know that it's a collection slice.

oed (commented):
Still not convinced inventing a new namespace for our json-schema usage is needed. Provided some more comments in this review.

I think we should represent CIP-82 using `ceramic://<streamid>` for single schemas and `ceramic://<streamid-1>,ceramic://<streamid-2>` for multiple schemas. If we need additional metadata like that this schema represents an appendCollection we can add a query: `ceramic://<streamid>?json-schema=appendCollection`

oed:
What's the status of this effort @PaulLeCam? I think an update is needed for https://github.com/ceramicnetwork/CIP/issues/87

PaulLeCam:
IIRC it should be good to merge, the main change that could be required is replacing the use of `tags` for AppendCollection but anyways the AppendCollection CIP will likely need other updates as we decide on an implementation.

oed (approved):
ok

PR #115, opened 2021-12-17
Does DocID need to be changed to StreamID in CiP-69?

bshambaugh:
This pull request does a direct string replacement of DociD with StreamID.


oed (approved):
Thanks!

PR #120, opened 2022-12-13
CIP: Multiple pubsub topics

oed:
[📚 Preview](https://github.com/ceramicnetwork/CIP/blob/cip/multiple-pubsub-topics/CIPs/cip-120.md)

### DIFF
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -18,12 +18,12 @@ Make the Ceramic protocol use multiple pubsub topics to gossip updates.
 ## Abstract
 <!--A short (~200 word) description of the technical issue being addressed.-->
 
-This CIP introduces the idea of separating into multiple pubsub topics based on a separator field in a streams genesis event, creating a unique pubsub topic for each value of this separator key. For streams that doesn't specify this field, the current pubsub topic is used for backwards compatibitliy.
+This CIP introduces the idea of separating into multiple pubsub topics based on a separator field in a streams genesis event, creating a unique pubsub topic for each value of this separator key. For streams that don't specify this field, the current pubsub topic is used for backwards compatibitliy.
 
 
 ## Motivation
 <!--Motivation is critical for CIPs that want to change the Ceramic protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.-->
-Currently Ceramic uses one pubsub topic to propagate *update*, *query*, and *response* messages in the network. As the network grows the scalability of this approach will be limited. Separating messages into multiple topics doesn't neccessarily solve this problem entirely. However, it allows developers of less popular streams to not share the burden message propagation for more scaled applications.
+Currently Ceramic uses one pubsub topic to propagate *update*, *query*, and *response* messages in the network. As the network grows the scalability of this approach will be limited. Separating messages into multiple topics doesn't neccessarily solve this problem entirely. However, it allows developers of less popular streams to not share the burden of message propagation for more scaled applications.
 
 
 ## Specification
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -3,7 +3,7 @@ cip: 120
 title: Multiple Pubsub Topics
 author: Joel Thorstensson <joel@3box.io>
 discussions-to: https://forum.ceramic.network/t/cip-120-multiple-pubsub-topics/
-status: Draft
+status: Last Call
 category: Standards
 type: Core
 created: 2022-12-13
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -73,7 +73,7 @@ Which is the pubsub topic currently used by Ceramic.
 
 ### Peer discovery
 
-While gossipsub used by Ceramic enables ambient discovery of peers, an initial set of peers already in a topic is required to recieve messages. It is therefore recommended that peers announce to the libp2p DHT that they are subscribed to a given pubsub topic. Peers can thus also discover other peers by looking up the DHT records for the given topic. The key that should be used for DHT providing and lookup MUST be `sha2-256(network-name || separator-key || separator-value)`.
+While gossipsub used by Ceramic enables ambient discovery of peers, an initial set of peers already in a topic is required to recieve messages. It is therefore recommended that peers announce to the libp2p DHT that they are subscribed to a given pubsub topic. Peers can thus also discover other peers by looking up the DHT records for the given topic. The key that should be used for DHT providing and lookup MUST be `sha2-256(topic-name)`.
 
 
 ## Rationale
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -73,7 +73,7 @@ Which is the pubsub topic currently used by Ceramic.
 
 ### Peer discovery
 
-While gossipsub used by Ceramic enables ambient discovery of peers, an initial set of peers already in a topic is required to recieve messages. It is therefore recommended that peers announce to the libp2p DHT that they are subscribed to a given pubsub topic. Peers can thus also discover other peers by looking up the DHT records for the given topic. The key that should be used for DHT providing and lookup MUST be `sha2-256(separator-key || separator-value)`.
+While gossipsub used by Ceramic enables ambient discovery of peers, an initial set of peers already in a topic is required to recieve messages. It is therefore recommended that peers announce to the libp2p DHT that they are subscribed to a given pubsub topic. Peers can thus also discover other peers by looking up the DHT records for the given topic. The key that should be used for DHT providing and lookup MUST be `sha2-256(network-name || separator-key || separator-value)`.
 
 
 ## Rationale
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -88,7 +88,7 @@ There are two approaches for how to transition to this new approach for pubsub t
 
 ### Duplicate messages
 
-For a transition period new nodes will publish updates in both the old main topic as well as the new topic. This ensures backwards compatibility in the system at the cost of some amount of time with, at worst, the amount of messages.
+For a transition period new nodes will publish updates in both the old main topic as well as the new topic. This ensures backwards compatibility in the system at the cost of some amount of time with, at worst, double the amount of messages.
 
 ### Force an upgrade (recommended)
 
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -49,7 +49,7 @@ If a stream has the following genesis header,
 {
   header: {
     sep: 'model',
-    model: <bytes-k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi>
+    model: toBytes('k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi', 'base36'))
   },
   ...
 }
--- a/CIPs/cip-120.md
+++ b/CIPs/cip-120.md
@@ -1,8 +1,8 @@
 ---
-cip: <to be assigned>
-title: <CIP title>
+cip: 120
+title: Multiple Pubsub Topics
 author: Joel Thorstensson <joel@3box.io>
-discussions-to: <URL of the Github issue for this CIP (if this is a PR)>
+discussions-to: https://forum.ceramic.network/t/cip-120-multiple-pubsub-topics/
 status: Draft
 category: Standards
 type: Core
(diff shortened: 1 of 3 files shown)


michaelsena:
lgtm 🔥

Michael Sena
Co-Founder, 3Box Labs

( https://3boxlabs.com/ )
https://3boxlabs.com ( https://3boxlabs.com )

michaelsena:
lgtm 🔥

michaelsena (approved):

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Simple Summary
+<!--Provide a simplified and layman-accessible explanation of the CIP.-->
+Make the Ceramic protocol use multiple pubsub topics to gossip updates.
+
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+
+This CIP introduces the idea of separating into multiple pubsub topics based on a separator field in a streams genesis event, creating a unique pubsub topic for each value of this separator key. For streams that doesn't specify this field, the current pubsub topic is used for backwards compatibitliy.
```
stbrody, line 21:
```suggestion
This CIP introduces the idea of separating into multiple pubsub topics based on a separator field in a streams genesis event, creating a unique pubsub topic for each value of this separator key. For streams that don't specify this field, the current pubsub topic is used for backwards compatibitliy.
```

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Motivation
+<!--Motivation is critical for CIPs that want to change the Ceramic protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.-->
+Currently Ceramic uses one pubsub topic to propagate *update*, *query*, and *response* messages in the network. As the network grows the scalability of this approach will be limited. Separating messages into multiple topics doesn't neccessarily solve this problem entirely. However, it allows developers of less popular streams to not share the burden message propagation for more scaled applications.
```
stbrody, line 26:
```suggestion
Currently Ceramic uses one pubsub topic to propagate *update*, *query*, and *response* messages in the network. As the network grows the scalability of this approach will be limited. Separating messages into multiple topics doesn't neccessarily solve this problem entirely. However, it allows developers of less popular streams to not share the burden of message propagation for more scaled applications.
```

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Example
+
+If a stream has the following genesis header,
+
+```javascript
+{
+  header: {
+    sep: 'model',
+    model: <bytes-k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi>
```
stbrody, line 52: What are we trying to convey with this `<bytes->` syntax?

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+```markdown
+/ceramic/<network-name>/
+```
+
+Which is the pubsub topic currently used by Ceramic.
+
+### Peer discovery
+
+While gossipsub used by Ceramic enables ambient discovery of peers, an initial set of peers already in a topic is required to recieve messages. It is therefore recommended that peers announce to the libp2p DHT that they are subscribed to a given pubsub topic. Peers can thus also discover other peers by looking up the DHT records for the given topic. The key that should be used for DHT providing and lookup MUST be `sha2-256(separator-key || separator-value)`.
```
stbrody, line 76: should the dht provide key have anything else to indicate it's for Ceramic specifically?  I guess the odds of someone random using the IPFS DHT to provide the hash of `"model || k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi"` is pretty low, but I still wonder if we should do more to namespace ourselves in the global DHT

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Backwards Compatibility
+<!--All CIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CIP must explain how the author proposes to deal with these incompatibilities. CIP submissions without a sufficient backwards compatibility section may be rejected outright.-->
+
+There are two approaches for how to transition to this new approach for pubsub topics.
+
+### Duplicate messages
+
+For a transition period new nodes will publish updates in both the old main topic as well as the new topic. This ensures backwards compatibility in the system at the cost of some amount of time with, at worst, the amount of messages.
```
stbrody, line 91: is the last sentence incomplete?  Is it meant to say at worst the amount of messages *doubled*?

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Force an upgrade (recommended)
+
+Almost like a hard fork we can pick a particular blockheight (since we track the eth blockchain), or blocktimestamp. Once this time has passed all nodes would switch over to the new pubsub topics (with maybe a short phase out period). The main drawback of this approach is that the community would need to coordinate to upgrade their software in time. There would be no backwards compatibility for nodes that didn’t make the upgrade.
```
stbrody, line 95: if we do this before composedb is supported on mainnet, and only do it for Model and MID streams, do we even need to do anything for backwards compatibility at all?

stbrody: I guess this also raises the question - do we want to do anything to TileDocument and Caip10Link streams?  Give them their own dedicated topic?  I guess we can't change the genesis contents without breaking determinism.  Makes me wonder if we should also encode the streamtype into the topic name, so streams with different streamtypes are always in a different pubsub topic.

stbrody (commented):
looking good.  

The proposal doesn't really give much guidance on how developers building on top of the base Ceramic protocol should think about using this new `sep` field.  It feels like it's at least worth a mention that there is some cost to having too many different pubsub topics, so a scheme like every stream having its own `sep` value would be a bad idea.

CIPs/cip-120.md
oed: That this is in fact not JSON, but we have the bytes that represents the streamid above.

CIPs/cip-120.md
oed: The streamid is already namespaced to Ceramic since it includes the `0xce` multicodec.

CIPs/cip-120.md
oed: I don't think we will introduce multiple pubsub topics before mainnet. Introducing the `sep` field I think we should do before mainnet, but this section talks more about the pubsub specific changes.

For Tile and Caip10Link I think we should not change anything, except that future Tiles can introduce the `sep`  field.

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Backwards Compatibility
+<!--All CIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The CIP must explain how the author proposes to deal with these incompatibilities. CIP submissions without a sufficient backwards compatibility section may be rejected outright.-->
+
+There are two approaches for how to transition to this new approach for pubsub topics.
+
+### Duplicate messages
+
+For a transition period new nodes will publish updates in both the old main topic as well as the new topic. This ensures backwards compatibility in the system at the cost of some amount of time with, at worst, the amount of messages.
```
oed, line 91:
```suggestion
For a transition period new nodes will publish updates in both the old main topic as well as the new topic. This ensures backwards compatibility in the system at the cost of some amount of time with, at worst, double the amount of messages.
```

CIPs/cip-120.md
oed: Made a change to add `network-name` so that different Ceramic networks don't overlap.

CIPs/cip-120.md
AaronGoldman: we could make it look not JSON by putting in javascript
```
base36.decode("k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi")
```

oed:
```suggestion
    model: toBytes('k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi', 'base36'))
```
makes sense!

CIPs/cip-120.md
stbrody: we can't add a `sep` field to tiles without breaking determinism.  What do you think about putting the streamtype into the topic name?  That would ensure that no Streams are using the top-level Ceramic topic, and let us keep that topic reserved for system-wide messages (like the keepalive messages used to share node versions, for example)

stbrody: I guess it does beg the question, are streamtypes something that the base event streaming layer should even be aware of?  At first glance it feels like that should be a higher-level concern, but it also kind of feels like it will be impossible for the base layer to avoid needing to know at least something about the streamtypes.  Especially as the type ID numbers are a global namespace and we'll need some registry of streamtypes to prevent collisions.

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
...
+```markdown
+/ceramic/mainnet/model/k2t6wyfsu4pg1hnttc39koxlw0pdn1kemzrn33cdfhrbooozvo7wb14sa1uqwi
+```
+
+### Streams without `sep`
+
+For backwards compatibility if a stream doesn't have a `sep` field or no value for the field specified by `sep`,  will use the following pubsub topic:
+
+```markdown
+/ceramic/<network-name>/
```
stbrody, line 69:
```suggestion
/ceramic/<network-name>/<streamid>
```

CIPs/cip-120.md
oed: > we can't add a sep field to tiles without breaking determinism

Not for existing apps, but for all new ones this wouldn't be a problem.

>  That would ensure that no Streams are using the top-level Ceramic topic, and let us keep that topic reserved for system-wide messages

If a node is only subscribed to one separator (e.g. model) I would assume they don't need to subscribe to a "top-level" topic, only the specific topic which they care about. It's probably a good idea to run keep alive messages on all subscribed topics for this reason. 

> but it also kind of feels like it will be impossible for the base layer to avoid needing to know at least something about the streamtypes

Would be curious to hear more about what you are thinking about this. 
Overall right now I don't see a strong reason to add streamtypes to the topic. Seems like a separate concern to me.

>  Especially as the type ID numbers are a global namespace and we'll need some registry of streamtypes to prevent collisions.

StreamType numbers are registered in the [StreamType table](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-59/CIP-59.md)

stbrody: > Not for existing apps, but for all new ones this wouldn't be a problem.

How would that work?  Each app would have to opt-in in their app code to using the new pubsub channel?  If we just changed the behavior in the TileDocument package it would affect all apps, old and new.

> If a node is only subscribed to one separator (e.g. model) I would assume they don't need to subscribe to a "top-level" topic

Maybe, maybe not.  I could definitely imagine many potential uses for a global topic that all Ceramic nodes subscribe to.  But I guess if that did become a need we could always create a new dedicated topic it for it rather than using the current topic.  We could leave the current topic to just be used by TileDocument and Caip10Link.  So maybe it's fine to leave streamtype out of the topic name.

> StreamType numbers are registered in the [StreamType table](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-59/CIP-59.md?rgh-link-date=2023-01-03T21%3A22%3A14Z)

That is a highly centralized approach.  I had assumed that at some point we'd want to move away from that to something more decentralized, and that whatever we do there might at least conceivably require Ceramic base layer support.  But I could also imagine solutions that don't require base layer support, so I think I agree that it makes sense to try to leave StreamTypes out of the base layer for now.

stbrody (approved):
lgtm

CIPs/cip-120.md
stbrody: revoking this proposal

CIPs/cip-120.md
```diff
@@ -0,0 +1,109 @@
+---
+cip: 120
+title: Multiple Pubsub Topics
+author: Joel Thorstensson <joel@3box.io>
+discussions-to: https://forum.ceramic.network/t/cip-120-multiple-pubsub-topics/
+status: Draft
```
oed, line 6:
```suggestion
status: Last Call
```

CIPs/cip-120.md
stbrody: only if they built their genesis commit manually.  The standard TileDocument APIs don't allow this: https://github.com/ceramicnetwork/js-ceramic/blob/b443590b6e346d5808e72d96fa868ec62114bff0/packages/stream-tile/src/tile-document.ts#L238

oed: Ok, this would be an easy code change though!

stbrody: Yep of course.  That's all I was trying to confirm with my original statement: app devs would have to manually opt-in to the new behavior.  We couldn't make the change for them.

PR #124, opened 2023-02-10
Recon

AaronGoldman:
A proposal synchronization of ranges of EventIDs from streams.


oed:
Please use `cip-124`

oed:
[📚 Preview](https://github.com/AaronGoldman/CIP/blob/stream-sets-synchronization/CIPs/CIP-124/cip-124.md)

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+
+We could partition on `model`, `controller`, `time` , or something else using a partition key? We could even map multiple dimensions into a single partition key and then partition based on the calulated partition key.
+
+What are the sorting orders that make sense?
+- Total time ordering of all events in all streams.
+    - `ceramic/<network>/<timeID>/<EventID>`
+- Sort by model/StreamID
+    - `ceramic/<network>/<model>/<EventID>`
+- Sort by controller/model/StreamID
+    - `ceramic/<network>/<did>/<model>/<EventID>`
```
oed, line 105:
```suggestion
    - `ceramic/<network>/<separator-key>/<separator-value-controller>/<separator-value>/<stream-controller><EventID>`
```

Compress by hash?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+  <multicodec-unsigned-varint><event hight>
+  <event-cid-bytes>
+```
+[multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint)
+
+#### **Stream set sort order**
+example lexicographical sort order
+```
+ceramic/<network>/<partitions_key>/<EventID>
+```
```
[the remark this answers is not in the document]
oed: If we want to include other orderings that we considered we should put those in rationale.

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+    - `ceramic/<network>/< z(model, controller, time) >`
+    - [https://youtu.be/YLVkITvF6KU](https://youtu.be/YLVkITvF6KU)
+    - [https://en.wikipedia.org/wiki/Z-order](https://en.wikipedia.org/wiki/Z-order)
+
+### **Interest Advertisements**
+In order to discover peers with overlaping interests nodes must avertise there interest ranges to the network.
+
+#### **DHT peer discovery**
+1. If known peer list too small
+    - Find providers in the DHT for the given StreamSetID `ceramic/<network>/<partition_key>/` (e.g. model)
```
oed, line 120: This key should use the same format as above.

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+    - Add the node to your known peer list if query receves a reponce.
+    - Advertise your node's stream set ranges of interests on the DHT.
+2. Send query to provider(s)
+    - Select peer(s) at random
+    - Maintain an internal reputation list for peers
```
oed, line 125: Should we define an algo for this?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+        - How up to date are they.
+        - Latency.
+        - Interests
+    - The query should include data about:
```
oed, line 129: Doesn't this spec need to define what this query looks like?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+        - The stream events you want.
+        - The stream events you have.
+        - If out of sync recur.
+3. Providers responds with their interests
```
oed, line 133: We should define this response structure as well?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+        - The stream events they have.
+        - The stream events they want.
+        - If out of sync recur.
+    - The node should add you to their known peer list after responding.
+    
+
+Periodically repeat the process above to achieve eventual consistency.
+
+```ts
+type InterestRange struct {
```
oed, line 144: What is this data structure?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+
+type Interests: InterestRange[]
+```
+
+#### **Interactive**
+One or more interest range,
+each containing one or more sub-tree associative hashs.
+```
+(EventID (sub-tree-ahash EventID)+ )+
+```
```
oed, line 167: What is this?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+      stop string 
+
+    }
+  ]
+}
+
+type Interests: InterestRange[]
+```
+
+#### **Interactive**
```
oed, line 162: Is this section supposed to describe the interactive algorithm between nodes? 
I like the use of animal names, but you should clarify that this is used instead of the long identifiers that were defined above.
Also feel like this section could be cleaned up a bit more to be easier to follow.

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+```js
+their_events: {"ape", "bee", "cat", "doe", "eel", "fox", "gnu", "hog"}
+your_events:  {"ape", "bee", "cat", "doe", "eel", "fox", "gnu", "hog"}
+
+you: -["ape", h("bee", "cat", "doe", "eel", "fox", "hog", "gnu"), "ape"]->
+
+```
+
+In this example only `"fox"` was never sent across the wire but in large sets with only tiny fractions of diffrences most keys will never be sent across the wire.
+> ToDo: should we merge
```
oed, line 217: Is this something that is discussed in the paper? Any advantage to this?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+> ["ape", h("bee", "cat", "doe", "eel", "fox", "hog", "gnu"), "ape"]
+> ```
+>  to
+> ```
+> [h("ape", "bee", "cat", "doe", "eel", "fox", "hog", "gnu")]
+> ```
+> to represent the entire ring?
+
+---
+#### **Imutable object store**
```
oed, line 228: This section is a bit too abstract for me to fully follow what you are getting at.

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+For syncing with an imutable object store add CIDs to the inodes. This would be DAG-JSON or DAG-CBOR
+```
+(EventID (sub-tree-ahash sub-tree-CID EventID)+ )+
+```
+This protocal can be run in a symalar way agenst a object store.
+however insted of the remote desiding which segments to send they pre-partion the set into a b-tree. At each laer of the b-tree you can deside if you have a match or if you need to recurse into the sub-tree.
+
+The use of synconising agenst a object store may improve the expereance of syncing a brand new node or a node that has been down for an extended pereod of time. The node could start by syncronising agenst a weekly snapshot and the sync with live nodes only after getting the bulk of the dataset from the object store.
+
+#### **DHT key**
```
oed, line 238: Isn't this covered in the **DHT peer discovery** section

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+
+### **Random Peer Synchronization Order**
+When synchronizing we want to sync with nearby nodes with higher probability since this is lower cost but also sometimes with the far away nodes so that the whole network is synchronized. By tracking the latency of other nodes with overlapping interests we can choose a peer proportional to the overlapping interests and inversely proportional to the latency.
+|| keys in overlapping interests || / latency
+
+![Random Gosip](assets/gossip.gif)
+
+The number of nodes with the event will grow exponetialy until it saturates the network.
+
+### **Associative Hash Function (SumOfShas)**
```
oed, line 251: Can this be moved to an appendix?

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+sub tree along with its max and min we only need to desend into that sub-tree if the reange end is in the sub-tree. Since a range has only two ends, start and stop, we can calulate the associative hash by visiting only twice the depth of the tree number of nodes `2*log_b(n)`
+where b is the fanout and n is the number of EventIDs.
+
+A b-tree with fanout 3:
+![fanout3](assets/b_hash_tree_1.png)
+
+A b-tree with fanout 2:
+![fanout2](assets/b_hash_tree_2.png)
+
+### **B#tree (B hash trees)**
```
oed, line 268: Not sure I follow the context of B-trees here. I didn't see them mentioned elsewhere?

oed (changes requested):
Seems like there's still some work on overall structure of this CIP needed. Right now it's a bit hard to read and understand how all pieces fit together.

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,321 @@
...
+  <genesis-cid-bytes>
+  <event-cid-bytes>
+
+// Proposed EventID
+<EventID> ::= <multibase-prefix><multicodec-streamid><stream-type>
+  <genesis-cid-bytes>
+  <multicodec-unsigned-varint><event hight>
+  <event-cid-bytes>
+```
+[multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint)
```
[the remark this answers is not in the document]
oed: Actually something that could make even more sense and would be less ambiguous:

```xml
<EventID> ::= <multibase-prefix><multicodec-streamid><stream-type><event-info>

<event-info> ::= <multicodec-eventid><truncated-streamid-bytes><varint event-height><event-cid>
```

This would allow us to more easily distinguish if it's an init event (just a CID multicodec) or event-id (with it's own multicodec)

CIPs/CIP-124/cip-124.md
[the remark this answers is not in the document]
oed: Updated the first comment to include the stream controller.

oed:
It might be nice to be able to reference a range as a `ceramic://...` uri

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,567 @@
...
+- Model
+    - `2/<Model Controller 6 bytes>/<StreamID full>/<Stream Hight [varint](https://github.com/multiformats/unsigned-varint)>/<Event CID full>`
+- Model Instance Document
+    - `3/<Model Controller 6 bytes>/<Model 6 bytes>/<Stream controller 6 bytes><StreamID full>/<Stream Hight [varint](https://github.com/multiformats/unsigned-varint)>/<Event CID full>`
+- UNLOADABLE
+    - Unloadable documents have no structure just use the StreamID as the key
+    - `4/<StreamID>`
+    - `4/kh4q0ozorrgaq2mezktnrmdwleo1d` only unloadable StreamID in [current use](https://www.notion.so/7a8a163fb97047359c56b26b6146ce85)
+
+The main risk in converting the events to keys is that we do want the keys to be grouped in ways that ceramic nodes are most likely to have large contiguous interests like the family for Tile Documents or the Model’s controller for Model Instance Documents. This is in tension with the desire to have short keys. Just the StreamID and event CID could be 124 characters add in an additional 6 char for `Model Controller`, `Model`, `Stream controller`, and 1 char for `Stream Hight` and `stream type` and we are talking about 142 chars for each stream.
```
AaronGoldman, line 207: Make clear that this is the `Stream controller` from the init event and not from replaying the complete stream.

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,567 @@
...
+- Z-order (Hilbert Curve order)
+    - `ceramic/<network>/< z(model, controller, time) >`
+    - advantages: 
+      - clustering by model, controller, and time
+    - disadvantages:
+      - complexity
+      - all three dimensions have few ranges but none are single range.
+    - notes:
+      - [https://youtu.be/YLVkITvF6KU](https://youtu.be/YLVkITvF6KU)
+      - [https://en.wikipedia.org/wiki/Z-order](https://en.wikipedia.org/wiki/Z-order)
```
qbig, line 426: wrong link, correct wiki page for Hilbert Curve
https://en.wikipedia.org/wiki/Z-order_(curve)

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,567 @@
...
+    - disadvantages: following a model or controller would be many ranges
+- Sort by model/StreamID
+    - `ceramic/<network>/<model>/<StreamID>/<event hight><Event CID>`
+    - advantages:
+      - following a models is a single range
+    - disadvantages:
+      - following a controllers is a many ranges
+
+- Sort by model_group/model/stream_group/StreamID/time/event
+    - `ceramic/<network>/<model controller>/<model>/<stream controller>/<StreamID>/<event hight>/<Event CID>`
```
qbig, line 413: <event hight> -> <event height>

CIPs/CIP-124/cip-124.md
```diff
@@ -0,0 +1,567 @@
+---
+cip: 124
+title: Recon
+author: Aaron Goldman (@aarongoldman) <aaron@3box.io>
+discussions-to: https://forum.ceramic.network/t/cip-124-recon-tip-synchronization-protocol/1144
+status: Draft
+category: Standards
+type: Networking
+created: 2023-01-18
+edited: 2023-01-18
```
oed, line 10:
```suggestion
edited: 2023-05-23
```

oed (approved):

CIPs/CIP-124/cip-124.md
oed: Let's update in a subsequent PR.

qbig:
@AaronGoldman @oed would using a bloom filter in addition to a hash during set recon make it more efficient (less round trips)? https://www.youtube.com/watch?v=xuddEiu-t-8&ab_channel=HeidiHoward

AaronGoldman:
fission's car-mirror work a little like that.
https://talk.fission.codes/t/notes-through-the-multiverse-ipfs-car-pool-car-mirror/3647

Certainly the split range function could avoid sending keys in the bloom filter as the splits.
This would reduce the likelihood of sending duplicate data at the cost of the bloom filter bandwidth.
Particularly as you get close to the leaves where there are not huge numbers of keys in the ranges it is likely to save the last few rounds. 

Screenshot 2023-05-23 at 3 13 43 PM

https://hur.st/bloomfilter/?n=&p=.001&m=2KiB&k=

e.g. If we through in a 2KiB filter we could probably send all remaining key when we got to about 1,000 keys in a range. Speeding up the last few rounds.

oed:
@qbig @AaronGoldman interesting discussion. Can we continue in the discussion thread for this CIP?
https://forum.ceramic.network/t/cip-124-recon-tip-synchronization-protocol/1144

PR #126, opened 2023-04-21
Update NFT DID to be purely generative.

oed:
This PR simplifies the NFT DID specification by a lot. This is achieved by making it purely generative. In order to use NFT DID going forward an object-capability in the form of a ChainProof would be needed to be used. This greatly reduces the complexity of the NFT DID overall since a Ceramic node would no longer need to rely on an external Blockchain indexer.

[📑 Preview](https://github.com/ceramicnetwork/CIP/blob/feat/update-nft-did/CIPs/cip-94.md)

### DIFF
--- a/CIPs/cip-94.md
+++ b/CIPs/cip-94.md
@@ -7,11 +7,12 @@ status: Draft
 category: Standards
 type: RFC
 created: 2021-02-12
+updated: 2023-04-21
 ---
 
 # NFT DID Method Specification
 
-The NFT DID Method converts any non-fungible token on any blockchain into a decentralized identifier where the owner of the NFT is the controller of the DID. This is achieved by using the *Chain Agnostic Improvement Proposals* to describe NFT assets and blockchain accounts, as well as the Ceramic network to find the DID associated with the owner of the NFT.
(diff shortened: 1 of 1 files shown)


stbrody (changes requested):
even though this doesn't require a blockchain indexer like TheGraph, it still requires a resolver of some kind to actually query the underlying blockchain state to validate the ChainProof right?  So for Ceramic to support NFT:DIDs for multiple different blockchains we'd still need logic to verify the ChainProofs for each supported blockchain, which requires a blockchain client and an API of some kind to query the underlying blockchain state?

oed:
@stbrody Given a ChainProof you would only need a blockchain light-client to verify its validity. I think NFT DID is useful beyond Ceramic, but in the case of Ceramic I could see us supporting Ethereum and all Eth L2s potentially.

oed:
Note to self: Would be useful to be able to put arbitrary delegations as part of the `tokenURL` metadata. This would only work if the metadata is generated from on-chain (for example, like [uniswap LP position NFTs](https://opensea.io/collection/uniswap-v3-positions-opt)). We could add a new field to the metadata object which could include a DID uri or an IPFS hash containing a capability object.

This would allow custom NFTs to be created where the owner can change delegations as they see fit. In particular this would be useful if a DAO want's to vote on who they delegate access to.

cc @ukstv you might find this interesting.

PR #127, opened 2023-04-28
CIP: CapReg - object-capability registry

oed:
[📜 Preview](https://github.com/ceramicnetwork/CIP/blob/cip/capreg/CIPs/CIP-1XX.md)


CIPs/cip-127.md
```diff
@@ -0,0 +1,207 @@
...
+Using a simple Ceramic stream, the capability registry enables users to notarize and revoke object-capabilities issued by their DID. The registry is based on hashes of object-capabilities encoded as CACAO.
+
+
+## Motivation
+
+<!--Motivation is critical for CIPs that want to change the Ceramic protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.-->
+
+Currently in Ceramic the main accounts types are PKH DID. These are great because they enable existing wallets to be used directly with Ceramic. Once a PKH DID is used to delegate permissions to a session key. That delegation will remain valid until the capability expires. This could be a problem in the case of a stolen session key or a malicious application. By introducing a capability registry these capabilities can be revoked at any time by the main DID (PKH DID in the case above, but this would work for any DID method). A system that uses these object capabilities could refer to the registry to verify that the capability  has not been revoked before it was used.
```
stbrody, line 28:
```suggestion
Currently in Ceramic the main accounts types are PKH DID. These are great because they enable existing wallets to be used directly with Ceramic. Once a PKH DID is used to delegate permissions to a session key, that delegation will remain valid until the capability expires. This could be a problem in the case of a stolen session key or a malicious application. By introducing a capability registry these capabilities can be revoked at any time by the main DID (PKH DID in the case above, but this would work for any DID method). A system that uses these object capabilities could refer to the registry to verify that the capability  has not been revoked before it was used.
```

CIPs/cip-127.md
```diff
@@ -0,0 +1,207 @@
...
+}
+```
+
+#### Streamid
+
+Generating the streamid can be done in three steps,
+
+1. Generate the multidid representation of the DID (see [Multidid specification]()).
+
+2. Encode the multidid as an inline CID,
```
stbrody, line 96: this is interesting as it breaks some assumptions in the Ceramic code as it exists today.  Currently we always assume the genesis CID can be loaded and used to create an initial StreamState object for the stream

CIPs/cip-127.md
```diff
@@ -0,0 +1,207 @@
...
+
+   ```solidity
```
stbrody, line 98: why is this marked as solidity code?

CIPs/cip-127.md
```diff
@@ -0,0 +1,207 @@
...
+  unixtime Integer // same as block.Time
+  event &TimestampEvent
+  block &EthereumHeader
+  path String // ipld path in the block to find event.txHash
+}
+```
+
+### Object Capabilities
+
+CapReg relies heavily on object-capabilities as they way to add and remove CACAO hashes in the registry.  Write access to CapReg can be delegated in the same way as delegting access to any other stream in Ceramic. Example using ReCap:
```
stbrody, line 166:
```suggestion
CapReg relies heavily on object-capabilities as the way to add and remove CACAO hashes in the registry.  Write access to CapReg can be delegated in the same way as delegting access to any other stream in Ceramic. Example using ReCap:
```

CIPs/cip-127.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Rationale
+
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+
+CapReg enables any DID to delegate full or partial permission to any other DID without having to worry about the capability getting lost since capabilities can now be revoked. The design requires some special logic for the state transition of the event stream to ensure that the capability is valid when the stream was updated. While Ceramic generally strives towards not including state transtion logic in event streams, for this particular case it seems difficult to avoid.
```
stbrody, line 185: > While Ceramic generally strives towards not including state transtion logic in event streams

What does this mean?

stbrody: don't all of our streams have state transition logic?

CIPs/cip-127.md
```diff
@@ -0,0 +1,207 @@
...
+The capability registry is based on a self-certifying data structure represented as a special type of Ceramic stream. Each DID has uniquely *one* capability registry. Any object capability issued by this DID can be notarized and revoked in the registry. Every update to the registry is recorded as a *Data Event*.
+
+```verilog
+type Prinicipal Bytes // multidid
+type CACAOHash Bytes // multihash
+type Varsig Bytes // varsig
+
+type Entry struct {
+  key CACAOHash 
+  revoked Boolean // true if the capability has been revoked
```
stbrody, line 47: it might be nice to make this more than a boolean, but rather a `state` string.  For instance at some point I think it could be very valuable to have a distinction between revoking a capability and invalidating it only after the time it is revoked, vs doing a revocation that also retroactively invalidates all writes that were ever done with the capability - decreeing the capability to never have been valid in the first place.  The latter could be useful if a user is tricked into signing a capability to a malicious app that then immediately starts publishing junk as the user.

Even if we only support "going forward" revocation in the first version, it would be nice to leave the door open to adding "retroactive" revocation as well in the future.

PR #135, opened 2023-05-28
Update old CIPs to appropriate statuses

oed:
It could be confusing for people to see multiple old CIPs being in review still. There's really no good reason to keep them there. They should either be made final and updated through subsequent CIP, or moved back to draft stage.

**Changes:**
* cip-5, streamtypes: review -> final
* cip-7, caip10-link: review -> final
* cip-8, tile document: review -> final
* cip-110, anchor contract: review -> final
* cip-120, multiple pubsub topics: review -> draft

### DIFF
--- a/CIPs/cip-110.md
+++ b/CIPs/cip-110.md
@@ -3,7 +3,7 @@ cip: 110
 title: Ceramic Anchor Contract
 author: Sergey Ukustov <sergey@ukstv.me>
 discussions-to: https://github.com/ceramicnetwork/CIP/issues/110
-status: Review
+status: Final
 category: RFC
 created: 2021-10-14
 edited: 2023-05-24
--- a/CIPs/cip-5.md
+++ b/CIPs/cip-5.md
@@ -3,7 +3,7 @@ cip: 5
 title: Streamtypes
 author: Michael Sena (@michaelsena), Joel Thorstensson (@oed), Janko Simonovic
 discussions-to: https://github.com/ceramicnetwork/CIP/issues/37
-status: Review
+status: Final
 category: RFC
 created: 2020-06-23
 edited: 2020-09-24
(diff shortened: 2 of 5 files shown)


CIPs/cip-110.md
```diff
@@ -3,7 +3,7 @@ cip: 110
 title: Ceramic Anchor Contract
 author: Sergey Ukustov <sergey@ukstv.me>
 discussions-to: https://github.com/ceramicnetwork/CIP/issues/110
-status: Review
+status: Final
```
stbrody, line 6: I think some of what's in this CIP doesn't match exactly what was built.  I assume once we mark something as Final we aren't supposed to change it, so should we update this CIP to be in alignment with what was built before we finalize it?

Some things I noticed:
1. The anchor event is called `DidAnchor` in this doc, but is actually called `anchorDagCbor` in reality.
2. The "Backwards Compatibility" section talks about adding `version: 1` to anchor commits, which we didn't do.  Instead we added `txType: "f(bytes32)"` to anchorProofs.

There may be more inconsistencies, I didn't check everything, that's just what I noticed from a quick scan

CIPs/cip-5.md
```diff
@@ -3,7 +3,7 @@ cip: 5
 title: Streamtypes
 author: Michael Sena (@michaelsena), Joel Thorstensson (@oed), Janko Simonovic
 discussions-to: https://github.com/ceramicnetwork/CIP/issues/37
-status: Review
+status: Final
```
stbrody, line 6: similar to my comment about the anchor contract CIP - there is still a lot of language in this CIP that uses "document" instead of "stream", and "Record" instead of "Commit".  Do we need to update the terminology to be consistent with what we use elsewhere before finalizing this?

oed:
@stbrody I can revert the changes on the two CIPs that you mentioned. Good otherwise?

CIPs/cip-110.md
oed: Yeah, we should update them to the extent that we can. Will take a pass here when I have time.

oed:
@stbrody I moved CIP-110 back to review. I also updated all other CIPs to use the correct language, so I think they should be good to move to final now.

PR #137, opened 2023-05-28
CIP: Ceramic API

oed:
This is an early draft of what could become the Ceramic API. An API that enables developers to read and write events directly to event streams.

### [Discussion forum thread](https://forum.ceramic.network/t/cip-137-ceramic-api/1153)

### DIFF
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -143,7 +143,7 @@ Get the raw IPLD data of a set of events given an array of eventids. This includ
 
 **Returns:**
 
-- *events* <string> - a base64 encoded CAR file containing the events
+- *events* <string> - a base64 encoded CAR file containing the event blocks
 
 ---
 
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -1,0 +1,200 @@
+---
+cip: 137
+title: Ceramic API
+author: Danny Browning <dbrowning@3box.io>, Nathaniel Cook <nathaniel@3box.io>, Aaron Goldman <aaron@3box.io>, Joel Thorstensson <joel@3box.io>
+discussions-to: <URL to a topic under the CIP category on the Ceramic forum: https://forum.ceramic.network/c/cips>
+status: Draft
+category: Interface
+created: 2023-05-28
+edited: 2023-05-28
+requires: 124
+---
+
+
+## Simple Summary
+<!--Provide a simplified and layman-accessible explanation of the CIP.-->
+An API for reading and writing events on Ceramic event streams.
+
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+TODO
+
+
+## Motivation
+<!--Motivation is critical for CIPs that want to change the Ceramic protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.-->
+TODO
+
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature.-->
+
+This is a work in progress specification that introduces a JSON-RPC interface for interacting directly with event streams on a node.
+
+
+
+### `ceramic_subscribe`
+
+This method tells the Ceramic node to subscribe to a given subset of streams. A stream subset is defined as `ceramic://*?<separator-key>=<separator-value>`, where *separator-key* describes a field in the header of the InitEvent of a stream,  and *separator-value* is the value of this field. For example to subscribe to all MIDs of a particular Model you can use `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`.
+
+Once subscribed the Ceramic node will use the Recon protocol to keep in sync with other nodes subscribed to this stream set.
+
+Active subscriptions need to be persisted. Or can be stateless somehow? (Ask Nathaniel)
+
+**Params:**
+
+- *ceramicUrl* <string> - the Ceramic URL to subscribe to
+- *options* <object> - options for the subscription
+    - *rangeNumber* <integer> - splits the stream set space into the given number of ranges
+    - *rangeIndex* <integer> - subscribes only to the given range index
+
+**Returns:**
+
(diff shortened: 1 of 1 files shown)


CIPs/cip-137.md
```diff
@@ -0,0 +1,200 @@
...
+
+**Returns:**
+
+- *range* <Array<string>> - a Recon message, representing the latest state of the range described by the subscription request. In the form of `[eventid, ahash, eventid]`
+
+---
+
+### `ceramic_unsubscribe`
+
+This method tells the Ceramic node to unsubscribe from a given stream set.
```
AaronGoldman, line 59: If there are multiple clients that want the same stream, do we want something like reference counting.

e.g.
If each subscription was given a UUID and we tracked the controller that asked for it a controller could delete their subscription without deleting the other subscriptions for the same range.

1) alice subscribes to `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`
2) bob subscribes to `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`
3) bob unsubscribes to `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`

at the end of 3 should the node be subscribed to `kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`?

having the controller and UUID would give the unsubscriber the option to delete only their unsubscription or all unsubscriptions for `kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`.

CIPs/cip-137.md
```diff
@@ -0,0 +1,200 @@
...
+
+**Returns:**
+
+- *reconRange* <Array<string>> - a Recon message, *`[eventid or ahash]`*
+
+---
+
+### `ceramic_getRawEvents`
+
+Get the raw IPLD data of a set of events given an array of eventids.
```
AaronGoldman, line 138: This is under specified about recursion.
If the event has a data link instead of a data Json should you pull that block also?
How deep should you go?

CIPs/cip-137.md
```diff
@@ -0,0 +1,200 @@
...
+
+- *[eventid]* <array<string>> - the events to fetch
+
+**Returns:**
+
+- *events* <string> - a base64 encoded CAR file containing the events
+
+---
+
+### `ceramic_addRawEvent`
```
AaronGoldman, line 150: would prefer `put` to imply the idempotency that http means by `put`

CIPs/cip-137.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+TODO
+
+
+## Motivation
+<!--Motivation is critical for CIPs that want to change the Ceramic protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the CIP solves. CIP submissions without sufficient motivation may be rejected outright.-->
+TODO
```
AaronGoldman, line 26: In order to enable databases other than ComposeDB to be built on top of ceramic we need a ceramic API for ceramic event streaming system to communicate with databases.

oed:
@AaronGoldman I addressed some of your comments, but can we please keep the normative discussions to the [forum post](https://forum.ceramic.network/t/cip-137-ceramic-api/1153)?

CIPs/cip-137.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature.-->
+
+This is a work in progress specification that introduces a JSON-RPC interface for interacting directly with event streams on a node.
+
+
+
+### `ceramic_subscribe`
```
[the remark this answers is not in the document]
oed: So what you are subscribing to is a separator-key/value, in the example it happens to be a model. It can definitely be possible to subscribe to specific streams within a separator!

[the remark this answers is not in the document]
qbig: Let's say stream-bar is from model-bar. So the "topic" could be "model-foo" or "stream-bar" or explicitly "model-foo/stream-bar"?

CIPs/cip-137.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+## Simple Summary
+<!--Provide a simplified and layman-accessible explanation of the CIP.-->
+An API for reading and writing events on Ceramic event streams.
+
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+This CIP introduces a JSON-RPC interface that can be used to subscribe to a number of streams inside of an interest range, and write new events to these streams. The API exposes two main ways of retrieving events. A simple method that uses the nodes local ordering of events to retrieve events by simply keeping track of a single vector client side. More advanced users can use a simplified version of the [Recon](https://cips.ceramic.network/CIPs/cip-124) protocol to retrieve events in their logical order.
```
AaronGoldman, line 21: why JSON-RPC over say GraphQL

CIPs/cip-137.md
```diff
@@ -0,0 +1,203 @@
...
+
+Get the raw IPLD data of a set of events given an array of eventids. This includes all IPLD blocks for this particular event, but no data from the previous events, e.g. DataEvents include signature envelope, event, and potentially detached payload, while TimeEvents include their entire merkle tree witness.
+
+**Params:**
+
+- *[eventid]* <array<string>> - the events to fetch
+
+**Returns:**
+
+- *events* <string> - a base64 encoded CAR file containing the events
```
AaronGoldman, line 146: events' blocks

AaronGoldman (approved):

CIPs/cip-137.md
oed: Sure, happy to change this if we have a good reason for it! Let's continue this discussion in the forum

CIPs/cip-137.md
oed, line 146:
```suggestion
- *events* <string> - a base64 encoded CAR file containing the event blocks
```

PR #140, opened 2023-06-11
CIP-137 update range spec

oed:
Based on a conversation with @AaronGoldman on how to best specify the sub-range of the subscription.

### DIFF
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -45,8 +45,8 @@ Active subscriptions need to be persisted. Or can be stateless somehow? (Nathani
 
 - *ceramicUrl* <string> - the Ceramic URL to subscribe to
 - *options* <object> - options for the subscription
-    - *rangeNumber* <integer> - splits the stream set space into the given number of ranges
-    - *rangeIndex* <integer> - subscribes only to the given range index
+    - *ctrlRangeStart* <string> - hex encoded string describing the start of the range of controllers to sync (defaults to `0x0000000000000000`)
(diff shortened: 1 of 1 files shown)


CIPs/cip-137.md
```diff
@@ -45,29 +45,36 @@ Active subscriptions need to be persisted. Or can be stateless somehow? (Nathani
 
 - *ceramicUrl* <string> - the Ceramic URL to subscribe to
 - *options* <object> - options for the subscription
-    - *rangeNumber* <integer> - splits the stream set space into the given number of ranges
-    - *rangeIndex* <integer> - subscribes only to the given range index
+    - *ctrlRangeStart* <string> - hex encoded string describing the start of the range of controllers to sync (defaults to `0x0000000000000000`)
```
AaronGoldman, line 48: Should we put the multibase `f` in front of this hex number?

for string that are not the full 64 bits I assume we are padding `ctrlRangeStart` with `0`s and `ctrlRangeEnd` with `f`s


I am assuming here that controller's range is referring to the controller by the first 64 bits of Sha256(DID string).
This may require some specification that the DID string includes the fragment and query parameters. `did:method:path#fragment`

If the DID string in the query does not match the DID string in the documents that is not a match even if they resolve to the same DID Document. We may want to warn somewhere if this is true.

oed: I think prefixing with `0x` is easier for developers since this is an existing convention that many use already.

> I am assuming here that controller's range is referring to the controller by the first 64 bits of Sha256(DID string).

I actually intended this to refer to the last 8 bytes of the sha256 hash, not the DID that is being hashed.

AaronGoldman (approved):
LGTM

PR #141, opened 2023-06-26
Feat: Ceramic API using OpenAPI

oed:
After some discussions with @stbrody it seems like using a REST api could be a good idea since it allows us to use multiple content types, among other benefits.

This PR proposes changing to using an http REST api. It also simplifies the api by reducing the number of endpoints needed.

### DIFF
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -57,6 +57,7 @@ This endpoint will respond as soon as the node knows about new events, given the
 * *eventOffset* <integer> - the number of events already consumed
 * *duration* <integer> - the amount of time, in seconds, to wait for a response
 
+
 #### Returns:
 
 There are two different content types supported as return value, as specified by the [`Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) http header.
@@ -157,6 +158,7 @@ This content type returns a json representation of the events requested. Note th
   - *header* <object> - header for the event
   - *data* <object> - the data of the event
   - *timestamp* <integer> - the unixtime this event was timestamped (if it has been)
+
 
 **CAR file:**
 
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -259,7 +259,7 @@ The Ceramic API provides a new way of interacting with event streams. The main b
 
 ## Implementation
 <!--The implementations must be completed before any CIP is given status "Final", but it need not be completed before the CIP is accepted.-->
-No implementation yet. Planned in [rust-ceramic](https://github.com/3box/rust-ceramic/)
+A partial implementation exists in [rust-ceramic](https://github.com/3box/rust-ceramic/). It currently implements the following [OpenAPI schema](https://github.com/3box/rust-ceramic/blob/main/api/ceramic.yaml).
 
 
 ## Security Considerations
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -59,7 +59,7 @@ This endpoint will respond as soon as the node knows about new events, given the
 
 #### Returns:
 
-There are two different content types supported as return value. 
+There are two different content types supported as return value, as specified by the [`Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) http header.
 
 **JSON:**
 
@@ -142,6 +142,8 @@ Retrieve events given an array of eventids.
 
 #### Returns:
 
+There are two different content types supported as return value, as specified by the [`Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) http header.
+
 **JSON:**
 
 Content-Type:  `application/json`
@@ -210,9 +212,11 @@ Only possible once we’ve migrated to use *Varsig* and can create events as pla
 
 #### Returns:
 
+
+
 **JSON:**
 
-- *success* <boolean> - true if the event was added correctly
+- *eventid* <string> - the eventId of this event
 
 
 
--- a/CIPs/cip-137.md
+++ b/CIPs/cip-137.md
@@ -1,7 +1,7 @@
 ---
 cip: 137
 title: Ceramic API
-author: Danny Browning <dbrowning@3box.io>, Nathaniel Cook <nathaniel@3box.io>, Aaron Goldman <aaron@3box.io>, Joel Thorstensson <joel@3box.io>
+author: Danny Browning <dbrowning@3box.io>, Nathaniel Cook <nathaniel@3box.io>, Aaron Goldman <aaron@3box.io>, Joel Thorstensson <joel@3box.io>, Spencer T Brody <spencer@3box.io> 
 discussions-to: https://forum.ceramic.network/t/cip-137-ceramic-api/
 status: Draft
 category: Interface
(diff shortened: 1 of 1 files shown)


CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
...
 
-**Params:**
 
-- *ceramicUrl* <string> - the Ceramic URL to unsubscribe from
-- *options* <object> - options for the subscription
-  - *ctrlRangeStart* <string> - hex encoded string describing the start of the range of controllers to sync (defaults to `0x0000000000000000`)
-  - *ctrlRangeEnd* <string> - hex encoded string describing the end of the range of controllers to sync (defaults to `0xffffffffffffffff`)
 
-**Returns:**
+**CAR file:**
```
dbcfd, line 81: Need a line return  here, otherwise the indenting is broken in the rest of the file when viewed.

CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
...
----
+* *{sort-key}* <string> - the *sort-key* and *sort-value* to subscribe to (e.g. `model` and `kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr` respectively)
+* *ctrlRangeStart* <string> - hex encoded string describing the start of the range of controllers to sync (defaults to `0x0000000000000000`)
+* *ctrlRangeEnd* <string> - hex encoded string describing the end of the range of controllers to sync (defaults to `0xffffffffffffffff`)
+* *eventOffset* <integer> - the number of events already consumed
+* *duration* <integer> - the amount of time, in seconds, to wait for a response
+
+#### Returns:
+
+There are two different content types supported as return value.
```
dbcfd, line 62: How is the content return specified? Query parameter or content-return header?

CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
...
-### `ceramic_recon`
+#### Example:
 
-Interact with the Ceramic node using the Recon protocol directly. This allows you to have greater control over the data you consume, but you have to be able to run the Recon algorithm client side.
+```
+/ceramic/events/kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr,kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr,kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr
+```
 
-**Params:**
+#### Returns:
```
dbcfd, line 143: Similar content above, how is return specified.

CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
...
-### `ceramic_putRawEvent`
+A possible convenience method for adding an event by only submitting a JWT.
 
-Add an event to a stream.
+Only possible once we’ve migrated to use *Varsig* and can create events as plain signed JWTs.
 
-**Params:**
+```javascript
+{
+  event: "<jwt>"
```
dbcfd, line 207: Can we describe this more? For example, which claims are required?

CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
...
+}
+```
 
-- *event* <string> - a base64 encoded CAR file containing the event
+#### Returns:
 
-**Returns:**
+**JSON:**
```
dbcfd, line 213: Does this need a return? Should this just be a 200 if successful, error otherwise?

CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
...
 ## Specification
 <!--The technical specification should describe the syntax and semantics of any new feature.-->
 
-This is a work in progress specification that introduces a JSON-RPC interface for interacting directly with event streams on a node.
+This specification introduces a REST API for interacting directly with event streams on a node.
 
 
 
-### `ceramic_subscribe`
+### GET `/ceramic/subscribe/{streamid}`
```
dbcfd, line 36: This should be a POST, even if we use query parameters for specifying the subscription.

dbcfd: A get to this endpoint would return information about the subscription.

dbcfd, line 36: Can we get an endpoint for getting all active subscriptions? Maybe `GET /ceramic/subscribe`?

dbcfd (changes requested):

CIPs/cip-137.md
oed: See `/stats/active-ranges`

oed: Not sure I follow. Why should it be a POST?

CIPs/cip-137.md
oed: That's a bug in the gitcoin renderer?

CIPs/cip-137.md
```diff
@@ -18,187 +18,227 @@ An API for reading and writing events on Ceramic event streams.
...
+This endpoint is called to subscribe to new events in a given interest range. This endpoint uses the node’s local ordering of events within the given range, constructed in the order in which the node received the events. If you want the global ordering of events you can use the *recon* endpoint. 
 
-Once subscribed the Ceramic node will use the Recon protocol to keep in sync with other nodes subscribed to this interest range.
+The range is normally defined as `ceramic://*?<sort-key>=<sort-value>`, where *sort-key* describes a field in the header of the InitEvent of a stream, and *sort-value* is the value of this field. For example to subscribe to all MIDs of a particular Model you can use `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`.
 
-Active subscriptions need to be persisted. Or can be stateless somehow? (Nathaniel to share more thoughts here)
+This endpoint will respond as soon as the node knows about new events, given the *eventOffset*, or when *duration* amount of seconds has passed.
```
nathanielc, line 42: Does this endpoint return a single event or multiple? What determines when we wait for more events? Is that what duration is for? 


For example lets say there is a stream that has events arriving about every 1s. If I call this endpoint with a duration of `5s` when do I get a response? Within 1s when the first event arrives or after 5s when my duration expires? or somewhere in between?  How would I make a call to this endpoint to get events with as little latency as possible? 

There is another challenge I see with using this endpoint in that `eventOffset` is the number of events. If we use number of events as the offset then the server must keep an index to each event so it can start anywhere in the stream and produce the event. In contrast the offset could be a byte offset into `log` file that the server produces. This would mean that when a client asks for new events the server can quickly seek to the correct place where the events are stored. This is how Kafka works as an example  see https://kafka.apache.org/documentation/#impl_reads

Kafka has two design differences that help keep the client simple and the server implementation manageable.

1. Clients asks for events from a byte offset
2. Clients specify a max size in bytes of events they are willing to receive.

The server then responds with as many events as can fit into the max bytes that are immediately available and returns the new offset that the clients can use on their next call. This allows for low latency when events are already present and allows the clients to control how much batching they need. 

Thoughts? Should we use some kind of byte offset instead of an event offset?

CIPs/cip-137.md
dbcfd: We're changing the state of the system, so should be one of POST/PUT/PATCH, with POST the best verb to use.

CIPs/cip-137.md
oed: Updated to return the eventId, as @PaulLeCam suggested.

CIPs/cip-137.md
oed: Depends on which system you are referring to. The node is simply just giving you information directly from the network. If it has this info locally already it will just give it to you, otherwise sync it from the network.

CIPs/cip-137.md
```diff
@@ -18,187 +18,223 @@ An API for reading and writing events on Ceramic event streams.
 
 ## Abstract
 <!--A short (~200 word) description of the technical issue being addressed.-->
-This CIP introduces a JSON-RPC interface that can be used to subscribe to a number of streams inside of an interest range, and write new events to these streams. The API exposes two main ways of retrieving events. A simple method that uses the nodes local ordering of events to retrieve events by simply keeping track of a single vector client side. More advanced users can use a simplified version of the [Recon](https://cips.ceramic.network/CIPs/cip-124) protocol to retrieve events in their logical order.
+This CIP introduces a REST API that can be used to subscribe to a number of streams inside of an interest range, and write new events to these streams. The API exposes two main ways of retrieving events. A simple method that uses the nodes local ordering of events to retrieve events by simply keeping track of a single vector client side. More advanced users can use a simplified version of the [Recon](https://cips.ceramic.network/CIPs/cip-124) protocol to retrieve events in their logical order.
```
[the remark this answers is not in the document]
oed: Agreed, we should add it as an asset associated with this CIP.

CIPs/cip-137.md
oed: Added more newlines, but that doesn't seem to fix the issue? Definitely seems like a bug.

dbcfd: Yeah, maybe try a horizontal line to separate?

oed: That kind of breaks the general flow though. Anyway this won't be a problem on cips.ceramic.network

CIPs/cip-137.md
nathanielc: > If we use a log file with byte offset, we would need to create a new log file for any new subscription (range).

This is the opposite of my intent with using the log file. We do not want the server to have to track any state for clients. Clients should be able to track all of their own state. My thinking is that the server would create a log file per unique sort-value. Then you need an offset per sort-value (i.e. an offset per model).  However that doesn't seem to work as we let ranges span multiple sort-values so you would need multiple offsets. Do we need to reconsider letting clients specify arbitrary ranges? Should we instead force clients to only be able to subscribe to a single sort-value at a time so that the server can manage unique sort-values efficiently? (This is similar to Kafka's partition concept, Kafka has two layers of namespacing, topic name and partition number). If we force a static hierarchy then the server can make tradeoffs accordingly. Need to think on this a bit more...

oed: > Should we instead force clients to only be able to subscribe to a single sort-value per subscription so that the server can manage unique sort-values efficiently?

This is not sufficient since we need to be able to subscribe to subset of sort-values if for example a model grows really big. One of the main value props of recon is that nodes can subscribe to arbitrary ranges.

> We do not want the server to have to track any state for clients.

Why is having one log file not considered state for clients? It's not like we would be keeping specific log files for specific clients, but rather for specific ranges.

nathanielc: >It's not like we would be keeping specific log files for specific clients, but rather for specific ranges.

There are an enumerable number of possible ranges so this means the server has to keep state based on the parameters of requests from clients. In the happy path there will be lots of reuse of ranges so its not an issue. But it leaves the door open for abuse as all a malicious client has to do is subscribe to lots of different ranges to cause the server to do lots of work. 


I think the core design question I am trying to answer is if we expect server implementations to maintain and index of each event or if they can reduce the overhead of events by storing them without an index.

I am going to write up a more complete discussion around this question.

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
-This is a work in progress specification that introduces a JSON-RPC interface for interacting directly with event streams on a node.
+This specification introduces a OpenAPI spec for interacting directly with event streams on a node.
 
 
 
-### `ceramic_subscribe`
+### GET `/ceramic/subscribe/{streamid}`
 
-This method tells the Ceramic node to subscribe to a given interest range of streams. A interest range is defined as `ceramic://*?<sort-key>=<sort-value>`, where *sort-key* describes a field in the header of the InitEvent of a stream, and *sort-value* is the value of this field. For example to subscribe to all MIDs of a particular Model you can use `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`.
+This endpoint is called to subscribe to new events in a given interest range. This endpoint uses the node’s local ordering of events within the given range, constructed in the order in which the node received the events. If you want the global ordering of events you can use the *recon* endpoint.
```
stbrody, line 38: it seems to me like a pretty shitty choice to either get events in an effectively random order or push all the processing and execution of a complex algorithm into the client.  Why can't there be an endpoint that both handles the Recon protocol for you and gives you events in the actual global order?

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
+This endpoint is called to subscribe to new events in a given interest range. This endpoint uses the node’s local ordering of events within the given range, constructed in the order in which the node received the events. If you want the global ordering of events you can use the *recon* endpoint. 
 
-Once subscribed the Ceramic node will use the Recon protocol to keep in sync with other nodes subscribed to this interest range.
+The range is normally defined as `ceramic://*?<sort-key>=<sort-value>`, where *sort-key* describes a field in the header of the InitEvent of a stream, and *sort-value* is the value of this field. For example to subscribe to all MIDs of a particular Model you can use `ceramic://*?model=kjzl6hvfrbw6c82mkud4qs38zl4hd03ifoyg2ksvfjkhuxebfzh3ef89vwvtvrr`.
 
-Active subscriptions need to be persisted. Or can be stateless somehow? (Nathaniel to share more thoughts here)
+This endpoint will respond as soon as the node knows about new events, given the *offset*, or when *duration* amount of seconds has passed. If the node already knows about a lot of new events, only *limit* amount of events will be returned.
```
stbrody, line 42: this is the first time the word "offset" appears in this document.  We should probably define what we're talking about here

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
----
 
 
+**CAR file:**
+
+Content-Type:  `application/vnd.ipld.car`
+
+This content types returns the underlaying IPLD data structure of the event streams encoded as a CAR file. This data can be used to trustlessly verify the integrity of the event stream. 
+
+The root object of the CAR file is a *SubscriptionResult* which contains an array of *EvenIds* the new *offset*. The CAR file also includes all *Event* objects, corresponding to the eventids in the subscription result.
```
stbrody, line 91:
```suggestion
The root object of the CAR file is a *SubscriptionResult* which contains an array of *EventIds* the new *offset*. The CAR file also includes all *Event* objects, corresponding to the EventIds in the subscription result.
```

stbrody: "[...] which contains an array of EventIds the new offset." <- the flow of this sentence doesn't make sense.  What are we trying to say about the new offset?

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
+```verilog
+// https://developers.ceramic.network/protocol/streams/event-log/#data-event
+type Event InitEvent | DataEvent | TimeEvent
+
+// https://cips.ceramic.network/CIPs/cip-124#eventids
+type EventId Bytes // sort-data + Event CID
+
+type SubscriptionResult {
+  events [EventId]
+	offset Integer
```
stbrody, line 104: is this supposed to be another indentation layer in?

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
+**CAR file:**
 
-### `ceramic_reconPoll`
+Content-Type:  `application/vnd.ipld.car`
 
-Same as `ceramic_recon`, but waits for *duration* amount of time in case any new events arrives at the node during this time.
+This content types returns the underlaying IPLD data structure of the event streams encoded as a CAR file. This data can be used to trustlessly verify the integrity of the event stream. 
 
-**Params:**
+The root object of the CAR file is *EventsResult* which is a convenience type containing an array of CIDs corresponding to the eventsids requested. The CAR file also includes all *Event* objects, including all IPLD blocks for these event, but no data from the previous events, e.g. DataEvents include signature envelope, event, and potentially detached payload, while TimeEvents include their entire merkle tree witness.
```
stbrody, line 172:
```suggestion
The root object of the CAR file is *EventsResult* which is a convenience type containing an array of CIDs corresponding to the EventsIds requested. The CAR file also includes all *Event* objects, including all IPLD blocks for these event, but no data from the previous events, e.g. DataEvents include signature envelope, event, and potentially detached payload, while TimeEvents include their entire merkle tree witness.
```

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
 
 
-### `ceramic_putEvent`
+---
 
-Convenience method for adding an event by only submitting a JWT.
+### GET `/stats/active-ranges`
 
-Only possible once we’ve migrated to use *Varsig* and can create events as plain signed JWTs.
+A status endpoint that returns all ranges that this node is actively monitoring.
```
stbrody, line 232: what if two clients send two subscription requests with overlapping ranges.  Do those show up here as 2 ranges or 1?

For example, client 1 requests range from A - E, client 2 requests range from C - G.  Do we report that as one range from A-G or as 2 ranges (A-E and C-G)?

CIPs/cip-137.md
```diff
@@ -18,190 +18,239 @@ An API for reading and writing events on Ceramic event streams.
...
 
-The choice of using json-rpc enables the API to work over multiple different transports. For example, HTTP, web sockets, or event the iframe postMessage api. The latter could be useful in the future if one want's to run a node inside of an iframe.
+The choice of using OpenAPI enables ease of integration and content types to more easily send and respond using binary formats such as CAR files. It's also a common pattern to have service workers serve http requests opening up for future browser native implementations of the Ceramic protocol.
 
 There are two main ways of polling for new events. Using recon and a more simple poll. This enables advanced developers to get full control with recon, while a simple poll API based on a single event counter is much easier to use for most developers. The disadvantage of the latter is that events will be received in the order that the node received them, not in the global order of the network.
```
stbrody, line 248: how is this easier for most developers?  How would a developer actually use this poll API?  How would **we** even use it?  If you get a newer event before an older one, it seems like most of the time you can't use that event at all until you get the earlier one anyway, so either you need to go out of band to fetch it, or wait for it to come in here.  Seems like it'd be better to just have the node do that for you

CIPs/cip-137.md
oed: How would that be possible? 
If the client consumes events from time `t` up to `t + n`, then the node hears about an event from time `t+1` over recon, what should the node do?

stbrody: within a given stream that shouldn't be possible.  Across streams we don't guarantee a total order anyway

CIPs/cip-137.md
stbrody: OOOOOHHHHH

Well yes that's a very different thing! That is not at all clear from this document however.  Yes, across streams events can come in any order.  But within a stream we should not return a newer event until we've received an earlier one.

stbrody: (mod events that come in on a different branch of history, which we'll still need to think about how we handle somehow)

oed: I think in this case clients would have to do consolidation as they get the events using the *eventHeight* from the EventId.

Fun side note: From a data structure perspective you could actually think of all MIDs that share the same Model as a single stream where the Model InitEvent is a kind of InitEvent for all MIDs.

AaronGoldman (approved):

PR #144, opened 2023-07-19
Feat/recon update sortkey

AaronGoldman:


CIPs/cip-124.md
```diff
@@ -66,10 +69,27 @@ Where:
 * `sort_value` is based on a user provided value for *sort-key* and *sort-value*
 * `controller` is the controller DID of the stream this event belongs to
 * `init_event_cid_bytes` is the CID of the first Event of the this stream.
-* `event_height` is the "height" of the event InitEvent. For InitEvents this value is `0` else `prev.event_height + 1`.
-* `event_cid_bytes` the CID of the event itself
+* `event_height` is the "height" of the event InitEvent. For InitEvents this value is `1` else `prev.event_height + 1`.
```
oed, line 72: I thought we were not making the change from `0` to `1`  on InitEvents?

oed (changes requested):
Should we add some language for "fencepost" in the ** Stream Set Ranges** section. Would also be good to update the examples there.

oed (approved):

PR #146, opened 2023-08-21
feat: after

AaronGoldman:
Allow an Event to indicate an EventID that it was created after.


smrz2001 (approved):
Merging so that further discussion can continue on the [forum](https://forum.ceramic.network/t/cip-146-after-discussion/1527) and in subsequent PRs.

PR #149, opened 2023-12-15
fix: pairing updates from forum discussion

smrz2001:


CIPs/cip-145.md
```diff
@@ -51,17 +56,15 @@ branches are pruned. Since pruning branches that contain Data Events would resul
...
-2. For branches that contain valid, uncovered Data Events, we only consider the first Data Event on a branch after the
-   fork.
-3. The Data Event that is covered by the earliest Time Event wins (see event `A` in fig. 5). 
-4. If two uncovered Data Events are covered by Time Events at the same block height, the Data Event with the lower CID
-   wins.
-
-Using multi-prev Data Events allows us to reduce the number of uncovered events and converge the stream so that there is
-only a single uncovered event, without any data abandoned on pruned branches. The stream's converged/diverged state can
-be determined by looking at the `prev` fields of all the Data Events for that stream.
+1. If a stream is in a diverged state (see events `A` and `B` in fig. 5), we consider branches that contain dominant
```
oed, line 59: Why would time events get pruned? 
**Consider Fig. 6:** 
* Data B expires at time `3<n<4`
* Data C has a time event at `4`
* If Time 3 is pruned, then it's unclear why Data B should remain valid

CIPs/cip-145.md
```diff
@@ -51,17 +56,15 @@ branches are pruned. Since pruning branches that contain Data Events would resul
...
+   Data Events. Branches containing only Time Events will get pruned.
+2. For branches that contain dominant Data Events, consider the first Data Event on a branch after the fork.
+3. The Data Event that is covered by the earliest Time Event wins (see event `A` in fig. 5).
+4. If two Data Events share the earliest timestamp, then the branch of the Data Event with the lower CID wins.
```
oed, line 63: What does it mean for a Data Event to "win" in the context of multiple prev?

smrz2001: We updated the wording. A Data Event "winning" here meant that that Data Event would become the tip.

oed: Ok I don't understand why we need to distinguish between winning and non-winning tips? Why not just call them all tips and put them all in the `prev` field? I don't follow why the protocol needs to care about "winning"?

CIPs/cip-145.md
m0ar: Good point. Since data A is the tip according to the rest of the rules in fig 4, I'm not sure why we should try to merge its descendants :thinking: 
image

CIPs/cip-145.md
oed: Yes, when I say protocol here I'm referring to the event streaming protocol. Stream type handlers is an application.

m0ar:
In general, these are the most important features from my perspective:
- All known commits, pruned or not, are pinned by the node
- A new node can learn about all known commits, pruned or not, from other nodes

The reason is that a historical shuffle should not prevent resolution of a once-valid commit, which we should be able to implement in the client using these two properties. Otherwise, there is an avenue for abusing late publish as an undo button. We would like to be able to rely heavily on deterministic resolution of state, and this is OK as long as the commits are preserved and communicated regardless of a consensus change.

If I understand this CIP correctly, I think the late publishing attack would be equated to just adding a new commit on the tip, which is anyway possible to do for the controller. What I'm not sure about though is what the merge of a divergent branch means for the state of the stream.

CIPs/cip-145.md
m0ar: Makes full-on sense then :ok_hand:

CIPs/cip-145.md
smrz2001: Yes, good question. @AaronGoldman and I had a good discussion about this scenario.
image
In this example, `Data C` can only be created after `Data B` has been validated to either be within the expiration timeout or to already have a valid Time Event.

We can assume that the author of `Data C` validated `Data B` because `Data C` follows `Data B`.

Of course, if `Time 3` for `Data B` is available before `Data C` is created, the CID of `Time 3` will be included in `Data C`'s multi-prev.

To your point, @m0ar, including `Time 1`'s descendants allows multi-prev to cover existing history (even if pruned) 	in the stream state, whereas today, pruned events are lost from the stream state.

oed: > We can assume that the author of Data C validated Data B because Data C follows Data B.

Can we really make this assumption? In theory the author of `Data C` could "import" `Data B` into the log even though it's invalid, e.g. `Time 3` doesn't exist or is too late.

smrz2001: > > We can assume that the author of Data C validated Data B because Data C follows Data B.
> 
> Can we really make this assumption? In theory the author of `Data C` could "import" `Data B` into the log even though it's invalid, e.g. `Time 3` doesn't exist or is too late.

Events like `Time 3` are definitely worth pinning, and potentially worth including in the multi-prev of subsequent Data Events.

We feel that the _validity_ of `Data C` does not require proof of validity of `Data A` or `Data B`. However, establishing the latter's _time bounds_ does require recording `Time 2` and `Time 3`. `Time 2` is already part of `Data C`'s history. Including `Time 3` in a future multi-prev would ensure that it also becomes part of the stream history.

oed: > We feel that the validity of `Data C` does not require proof of validity of `Data A` or `Data B`. However, establishing the latter's time bounds does require recording `Time 2` and `Time 3`. `Time 2` is already part of `Data C`'s history. Including `Time 3` in a future multi-prev would ensure that it also becomes part of the stream history.

Agree that `Data C`'s validity is not predicated on the validity of `A` or `B`. However, the validity of `Data B` **is** predicated on `Time 3` being available. Therefore, it doesn't seem right to say that we can prune `Time 3`.

smrz2001: > > We feel that the validity of `Data C` does not require proof of validity of `Data A` or `Data B`. However, establishing the latter's time bounds does require recording `Time 2` and `Time 3`. `Time 2` is already part of `Data C`'s history. Including `Time 3` in a future multi-prev would ensure that it also becomes part of the stream history.
> 
> Agree that `Data C`'s validity is not predicated on the validity of `A` or `B`. However, the validity of `Data B` **is** predicated on `Time 3` being available. Therefore, it doesn't seem right to say that we can prune `Time 3`.

Yes, what would you think about including `Time 3` in the multi-prev of the next event to be added to the stream? That's what we meant to say here:
> Events like Time 3 are definitely worth pinning, and potentially worth including in the multi-prev of subsequent Data Events.

What if a new event `Data D` (occurring after `Time 4`) had a `prev` of `[Time 4, Time 3]`? Even though `Time 3` will not take precedence during tip selection, it will always remain part of the stream history.

Or, let's say, `Time 3` didn't show up until we already had `Time 4` -> `Data D` -> `Time 5` -> `Data E` -> `Time 6`. Then `Data F` (occurring after `Time 6`) would have a `prev` of `[Time 6, Time 3]`. `Data B` would remain in an "unverified" state until `Time 3` was discovered.

Tracking `Data B`'s validity this way would be a little more complicated than the usual flow, but always possible. Moreover, now _all_ events related to the stream would be part of the DAG, which is great.

oed: Yes, this is what I mean! Just wanted to be clear that `Time 3` can't be pruned if we expect the creator of `Data D` to include it in *prev*.

CIPs/cip-145.md
smrz2001: > Ok I don't understand why we need to distinguish between winning and non-winning tips? Why not just call them all tips and put them all in the `prev` field? I don't follow why the protocol needs to care about "winning"?
> e.g. The protocol just gives the application a list of tips. It's up to the application to decide what is "winning" and what's not. It's also up to the application to chose the order of tips in it's `prev` array when it's being constructed.

To answer your question, @oed, there are two reasons for this:
- It is simpler for applications to just be given a tip per some default, predictable algorithm. They can choose to override this order, but don't have to.
- There is always _some_ eventually consistent state of a diverged stream, even if the controller never comes back to create the Merge Event, because the default precedence rules can be used to determine the tip.

CIPs/cip-145.md
```diff
@@ -51,17 +59,33 @@ branches are pruned. Since pruning branches that contain Data Events would resul
...
+1. Based on rule (1), the stream state has 4 candidate tips, `Time 5`, `Time 6`, `Time 4`, and `Data F`. One of these
+   candidate tips will become the tip of the stream.
+   ![Alt text](../assets/cip-145/rules2.png)
+2. The stream forks between the branches for `Time 1` and `Data A`. Based on rules (3) and (4), the branch for `Data A`
+   is the only branch considered for tip selection.
+   ![Alt text](../assets/cip-145/rules3.png)
+3. This branch later forks into additional branches for `Data E`, `Time 4`, and `Data F`. Based on rules (3) and (4),
+   the branches for `Data E` and `Data F` are the only branches considered for tip selection.
+   ![Alt text](../assets/cip-145/rules4.png)
+4. Based on rule (5), since there is a Time Event for `Data E` but not for `Data F`, only the branch for `Data E` is
```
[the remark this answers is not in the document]
smrz2001: Good catch, yes, it should be rule 4. It is time that made the decision because a Data Event without a Time Event is as if it occurred at time infinity.

> Can we end up in the case where there is a yet-unknown, earlier anchor for the other branch in transit?

Yes, that's possible. If there was an as of yet unknown `Time 5.5` corresponding to `Data F`, then `Time 5.5` becomes the tip.

While a lot less likely in the absence of a malicious CAS attempting a late-publishing attack, it is also possible for example for a `Time 1.5` covering `Data B` to be discovered late. This would rewind the state of the stream, marking `Time 5` the tip.

Having said that, this spec provides a way for the application to resolve such a situation without data loss. A user can decide whether to override the default tip with a new event, while keeping the stream history intact.

CIPs/cip-145.md
```diff
@@ -51,17 +59,33 @@ branches are pruned. Since pruning branches that contain Data Events would resul
...
-4. If two uncovered Data Events are covered by Time Events at the same block height, the Data Event with the lower CID
-   wins.
-
-Using multi-prev Data Events allows us to reduce the number of uncovered events and converge the stream so that there is
-only a single uncovered event, without any data abandoned on pruned branches. The stream's converged/diverged state can
-be determined by looking at the `prev` fields of all the Data Events for that stream.
+1. If a stream is in a diverged state, each uncovered event is a candidate tip.
+2. Branches that do not contain dominant Data Events cannot be the tip.
+3. For branches that contain dominant Data Events, consider the earliest Data Event after a fork point.
+4. The branch with the earliest Data Event becomes the tip.
```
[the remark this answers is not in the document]
smrz2001: Yes, the anchored one is earlier by definition. A Data Event without a Time Event is as if it occurred at time infinity.

We'll update the rule to state this.

CIPs/cip-145.md
oed: Wouldn't the application need to be given all tips anyway? In order to include them all in the `prev` array? Are we simply talking about the ordering of the CIDs in the returned array here?

CIPs/cip-145.md
oed: @smrz2001 since it seems hat we are in agreement, maybe you can update the language to be clear that the TimeEvent doesn't get pruned?

CIPs/cip-145.md
```diff
@@ -51,17 +59,33 @@ branches are pruned. Since pruning branches that contain Data Events would resul
...
-3. The Data Event that is covered by the earliest Time Event wins (see event `A` in fig. 5). 
-4. If two uncovered Data Events are covered by Time Events at the same block height, the Data Event with the lower CID
-   wins.
-
-Using multi-prev Data Events allows us to reduce the number of uncovered events and converge the stream so that there is
-only a single uncovered event, without any data abandoned on pruned branches. The stream's converged/diverged state can
-be determined by looking at the `prev` fields of all the Data Events for that stream.
+1. If a stream is in a diverged state, each uncovered event is a candidate tip.
+2. Branches that do not contain dominant Data Events cannot be the tip.
+3. For branches that contain dominant Data Events, consider the earliest Data Event after a fork point.
```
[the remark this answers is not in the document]
smrz2001: Yes, we actually did mean to refer to anchor time here. This helps keep the language consistent with other places.

We hope this earlier clarification of a `fork point` help clarify what we mean:
```
* A `fork point` for a branch is the earliest event on that branch that is not on another branch. 
```

smrz2001:
> In general, these are the most important features from my perspective:
> 
> * All known commits, pruned or not, are pinned by the node
> * A new node can learn about all known commits, pruned or not, from other nodes
> 
> The reason is that a historical shuffle should not prevent resolution of a once-valid commit, which we should be able to implement in the client using these two properties. Otherwise, there is an avenue for abusing late publish as an undo button. We would like to be able to rely heavily on deterministic resolution of state, and this is OK as long as the commits are preserved and communicated regardless of a consensus change.
> 
> If I understand this CIP correctly, I think the scope of a late publishing attack would be equated to just adding a new commit on the tip, which is anyway possible to do for the controller. What I'm not sure about though is what the merge of a divergent branch means for the state of the stream, but I'm not sure if this is relevant in this context.

The two big advantages of this CIP are:
- The tip event is a discovery mechanism for _and commitment to_ not only the prev-chain but the prev-DAG.
- You can at any point in time account for which branches are pruned and which have been merged.

